PR #353, opened 2022-10-12
Refactor newmessageview

PhilippeWeidmann:


PhilippeWeidmann:
Partially fix #340
Message was already downloaded before giving it to the ComposeMessageView but body wasn't updated.
As this is a rather poor experience we will need to decide between this 4 options:
- Show nothing to the user when replying from thread list and downloading the body
- Show a loading indicator in the quick actionview
- Show a loading indicator in the ComposeMessageView
- Wait for the "New api" as we will only handle messages we will always have message body

Lnamw (approved):

Ambrdctr (approved):

PR #442, opened 2022-12-21
Refactor draft

PhilippeWeidmann:


Ambrdctr (approved):

MailCore/Models/Draft.swift
```diff
@@ -482,7 +185,120 @@ public class Draft: Object, Decodable, Identifiable, AbstractDraft {
         self.action = action
     }
 
-    public func asUnmanaged() -> UnmanagedDraft {
-        return .toUnmanaged(managedDraft: self)
+    public static func mailTo(subject: String?,
+                              body: String?,
+                              to: [Recipient], cc: [Recipient], bcc: [Recipient]) -> Draft {
```
adrien-coye, line 190: If we put arguments under each other I would do it for all of them.
I see the Aestetics of keeping to:cc:bcc together, so maybe we should wrap in into a `MailRecipients` struct ?

MailCore/Models/Draft.swift
```diff
@@ -482,7 +185,120 @@ public class Draft: Object, Decodable, Identifiable, AbstractDraft {
...
+                subject = "Fwd: \(subject)"
+            }
+            quote = Constants.forwardQuote(message: message)
+            attachments = attachmentsToForward
+        }
+
+        var to: [Recipient] = []
+        var cc: [Recipient] = []
+
+        if mode.isReply {
```
adrien-coye, line 223: If we start writing some more unit tests, in the future keeping a dedicated func for the isReply case could be usefull.
That would mean this replying(to:) func would only be here to dispatch the correct specialized func.

MailCore/Models/Draft.swift
```diff
@@ -482,7 +185,120 @@ public class Draft: Object, Decodable, Identifiable, AbstractDraft {
...
+            let userEmail = AccountManager.instance.currentMailboxManager?.mailbox.email ?? ""
+            let cleanedFrom = Array(message.from.detached()).filter { $0.email != userEmail }
+            let cleanedTo = Array(message.to.detached()).filter { $0.email != userEmail }
+            let cleanedReplyTo = Array(message.replyTo.detached()).filter { $0.email != userEmail }
+            let cleanedCc = Array(message.cc.detached()).filter { $0.email != userEmail }
+
+            to = cleanedReplyTo.isEmpty ? cleanedFrom : cleanedReplyTo
```
adrien-coye, line 230: It looks like an unrolled loop, probably easier to review than a `while(to.isEmpty) { }.`
I can't think of something easier to read ATM.

adrien-coye (commented):
I had some questions to try to get a better understanding of the project.

Mail/Views/New Message/ComposeMessageView.swift
```diff
@@ -267,6 +248,18 @@ struct ComposeMessageView: View {
                 EmptyView()
             }
         }
+        .task {
+            if draft.messageUid != nil && draft.remoteUUID.isEmpty {
```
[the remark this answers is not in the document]
valentinperignon: This way of thinking felt a bit unnatural at the beginning for me, but I am starting to have the reflex to put guards! 😄

Mail/Components/RecipientField.swift
```diff
@@ -41,7 +42,7 @@ struct RecipientChip: View {
 }
 
 struct RecipientField: View {
-    @Binding var recipients: [Recipient]
+    @Binding var recipients: RealmSwift.List<Recipient>
```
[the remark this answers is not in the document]
PhilippeWeidmann: We use a realm list because we are binding to an object directly stored into realm `Draft.to: List<Recipient>` same for cc and bcc

Mail/Utils/DraftUtils.swift
```diff
@@ -21,26 +21,23 @@ import MailCore
 import SwiftUI
 
 class DraftUtils {
-    @MainActor public static func editDraft(from thread: Thread, mailboxManager: MailboxManager, editedMessageDraft: Binding<Draft?>) {
+    public static func editDraft(from thread: Thread, mailboxManager: MailboxManager, editedMessageDraft: Binding<Draft?>) {
```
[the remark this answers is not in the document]
PhilippeWeidmann: It's still not available for execution outside main thread but we are not using it from outside the main thread anywhere.
Still this is not good for future method consumer who won't be aware of this, I will add it back to the whole class.

Mail/Views/New Message/ComposeMessageView.swift
```diff
@@ -154,10 +151,10 @@ struct ComposeMessageView: View {
                                 .focused($focusedField, equals: .subject)
                         }
 
-                        if let attachments = draft.attachments?.filter { $0.contentId == nil }, !attachments.isEmpty {
+                        if let attachments = draft.attachments.filter { $0.contentId == nil }, !attachments.isEmpty {
                             ScrollView(.horizontal, showsIndicators: false) {
                                 HStack(spacing: 8) {
-                                    ForEach(attachments) { attachment in
+                                    ForEach(Array(attachments)) { attachment in
```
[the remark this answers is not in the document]
adrien-coye: like discussed maybe a .toArray() extension

MailCore/Models/Draft.swift
adrien-coye: discussed

MailCore/Models/Draft.swift
adrien-coye: discussed

Ambrdctr (approved):

Ambrdctr (approved):

Lnamw (approved):

PR #463, opened 2022-12-22
Change color illustration

valentinperignon:
Depends on #441


adrien-coye (commented):
Can we try to update the Lottie package ? It probably will not solve the freezing issue, but looking at the code I can only think of this.

Lnamw (dismissed):

Ambrdctr (approved):

Ambrdctr (approved):

Lnamw (approved):

PR #467, opened 2023-01-02
fix: Rotating device between normal orientation and flat

Ambrdctr:


Mail/Utils/View+Extension.swift
```diff
@@ -23,12 +23,15 @@ import SwiftUI
 
 struct DeviceRotationViewModifier: ViewModifier {
     let action: (UIInterfaceOrientation?) -> Void
+    @State var lastOrientation = UIApplication.shared.mainSceneKeyWindow?.windowScene?.interfaceOrientation
```
adrien-coye, line 26: I just have a question, could we use a binding instead of an internal state to handle rotation ?

adrien-coye: We checked together quickly this morning with @Ambrdctr it was not helping or easier to read.

valentinperignon (approved):

adrien-coye (approved):
All good

PR #469, opened 2023-01-05
fix: UnreadCount

Ambrdctr:


MailCore/Models/Folder.swift
```diff
@@ -178,8 +178,7 @@ public class Folder: Object, Codable, Comparable, Identifiable {
     }
 
     public func computeUnreadCount(using realm: Realm) {
```
PhilippeWeidmann, line 180: realm isn't needed anymore

MailCore/Models/Folder.swift
```diff
@@ -178,8 +178,7 @@ public class Folder: Object, Codable, Comparable, Identifiable {
...
-        let unreadMessages = realm.objects(Message.self).where { $0.folderId == self.id && $0.seen == false }
-        unreadCount = unreadMessages.count
+        unreadCount = threads.where { $0.hasUnseenMessages == true }.count
```
PhilippeWeidmann, line 181: This seems wrong, if I have a thread with 2 messages marked unread it will still display 1 unread in the header ?

PhilippeWeidmann (changes requested):

MailCore/Models/Folder.swift
adrien-coye: Yes it looks like this counts the number of threads with unread messages in them, is this what is expected?

PhilippeWeidmann (approved):

PR #493, opened 2023-01-12
fix: Filter message references

Ambrdctr:


adrien-coye (commented):
I'll be back to re review with the regex.

MailCore/Models/Message.swift
```diff
@@ -166,20 +166,12 @@ public class Message: Object, Decodable, Identifiable {
...
 
-    public func computeReference() {
-        if var refs = references, !refs.isEmpty {
-            refs.removeFirst()
-            refs.removeLast()
-            refs = refs.replacingOccurrences(of: "> <", with: "><")
-            let refsArray = refs.components(separatedBy: "><")
-            linkedUids.insert(objectsIn: refsArray)
+    public func computeReference(using regex: NSRegularExpression) {
+        if let refs = references, !refs.isEmpty {
```
PhilippeWeidmann, line 170: We can simply unwrap using `if let references`

MailCore/Cache/MailboxManager.swift
```diff
@@ -694,28 +694,39 @@ public class MailboxManager: ObservableObject {
...
-                    message.inTrash = folder.role == .trash
-                    message.computeReference()
-                    let existingThreads = Array(realm.objects(Thread.self)
-                        .where { $0.messageIds.containsAny(in: message.linkedUids) })
-
-                    if let newThread = createNewThreadIfRequired(for: message, folder: folder, existingThreads: existingThreads) {
-                        threadsToUpdate.insert(newThread)
-                    }
+        do {
+            let regex = try NSRegularExpression(pattern: Constants.referenceRegex)
```
PhilippeWeidmann, line 698: Place the whole NSRegularExpression in a static constant to avoid recreating it for each thread

MailCore/Utils/Constants.swift
```diff
@@ -96,6 +96,7 @@ public enum Constants {
 
     public static let mailRegex =
         "(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|\"(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21\\x23-\\x5b\\x5d-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])*\")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21-\\x5a\\x53-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])+)\\])"
+    public static let referenceRegex = ">\\s*<|>?\\s+<?"
```
PhilippeWeidmann, line 99: Compute the NSRegularExpression only once (and remove ...using(regex:) where needed)
```
    public static var referenceRegex: NSRegularExpression = {
        // It's safe to unwrap because this will always succeed
        // swiftlint:disable force_try
        return try! NSRegularExpression(pattern: ">\\s*<|>?\\s+<?")
    }()
    ```

PhilippeWeidmann (changes requested):

PhilippeWeidmann (approved):

Lnamw (approved):

PR #522, opened 2023-01-23
fix: Inline shown as attachment if cid is empty

Ambrdctr:


PhilippeWeidmann (approved):

Mail/Views/Thread/MessageView.swift
```diff
@@ -50,7 +50,7 @@ struct MessageView: View {
                 .padding(.horizontal, 16)
 
                 if isMessageExpanded {
-                    if !message.attachments.filter { $0.contentId == nil }.isEmpty {
+                    if !message.attachments.filter { $0.disposition == .attachment || $0.contentId == nil }.isEmpty {
```
valentinperignon, line 53: We have to filter the attachments twice : in AttachmentsView and in MessageView.
Is it possible to refactor a bit to filter only once?

Maybe it's beyond your scope and could be done later, it's just a suggestion.

valentinperignon (approved):

PR #532, opened 2023-01-30
refactor: Custom alerts as fullscreen sheet

PhilippeWeidmann:
Present alerts fullscreen anywhere.
Present alert from an object binding instead of a Bool.

PR #534, opened 2023-01-31
feat: Attachment global errors

PhilippeWeidmann:
Depends on #532

### DIFF
--- a/Mail/Views/New Message/Attachments/AttachmentsManager.swift
+++ b/Mail/Views/New Message/Attachments/AttachmentsManager.swift
@@ -31,6 +31,7 @@ class AttachmentUploadTask: ObservableObject {
     }
 }
 
+@MainActor
 class AttachmentsManager: ObservableObject {
     private let draft: Draft
     private let mailboxManager: MailboxManager
@@ -43,6 +44,8 @@ class AttachmentsManager: ObservableObject {
         return attachmentUploadTasks.values.allSatisfy(\.uploadDone)
     }
 
+    var globalError: MailError?
+
     private lazy var filenameDateFormatter: DateFormatter = {
         let formatter = DateFormatter()
         formatter.dateFormat = "yyyyMMdd_HHmmssSS"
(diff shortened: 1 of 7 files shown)


Mail/Views/AlertView.swift
```diff
@@ -47,23 +61,69 @@ struct CustomAlertModifier<AlertContent>: ViewModifier where AlertContent: View
...
 
+private struct ClearFullScreenView: UIViewRepresentable {
+    private class BackgroundRemovalView: UIView {
+        override func didMoveToWindow() {
+            super.didMoveToWindow()
+            clearBackgroundSuperviews(view: self)
+        }
+
+        private func clearBackgroundSuperviews(view: UIView, level: Int = 0) {
+            guard level < 5 else {
```
adrien-coye, line 101: Can we name the const to make more explicit what it does ?

Mail/Views/AlertView.swift
```diff
@@ -47,23 +61,69 @@ struct CustomAlertModifier<AlertContent>: ViewModifier where AlertContent: View
...
+
+            if let superview = view.superview {
+                superview.backgroundColor = .clear
+                clearBackgroundSuperviews(view: superview, level: level + 1)
+            }
+        }
+    }
+
+    func makeUIView(context: Context) -> UIView {
+        return BackgroundRemovalView()
```
adrien-coye, line 113: You are more into returning directly than I might be, but what is the prefered approach with implicit returns in closures ? I like to ommit retrun when possible.

Mail/Views/New Message/Attachments/AttachmentsManager.swift
```diff
@@ -26,8 +26,12 @@ class AttachmentUploadTask: ObservableObject {
     @Published var progress: Double = 0
     var task: Task<String?, Never>?
     @Published var error: MailError?
+    var uploadDone: Bool {
+        return progress == 1
```
adrien-coye, line 30: It does not change anything, but I'd write >= 1.

adrien-coye (commented):
I only had small questions. LGTM.

Mail/Views/AlertView.swift
PhilippeWeidmann: Resolved in dependant PR

Lnamw (dismissed):

Ambrdctr (approved):

Lnamw (approved):

PR #549, opened 2023-02-03
feat: Inject KeychainHelper from Core

PhilippeWeidmann:


Mail/AppDelegate.swift
```diff
@@ -106,12 +106,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let networkLoginService = Factory(type: InfomaniakNetworkLogin.self) { _, _ in
```
adrien-coye, line 106: This is a nice to have, but while injecting objects, we might want to register `InfomaniakNetworkLoginable.self` here.
That way it will be easier to test.

MailCore/API/MailApiFetcher.swift
```diff
@@ -291,18 +291,19 @@ public class MailApiFetcher: ApiFetcher {
...
 class SyncedAuthenticator: OAuthAuthenticator {
-    @InjectService var networkLoginService: InfomaniakLogin
-
     override func refresh(
         _ credential: OAuthAuthenticator.Credential,
         for session: Session,
         completion: @escaping (Result<OAuthAuthenticator.Credential, Error>) -> Void
     ) {
         AccountManager.instance.refreshTokenLockedQueue.async {
+            @InjectService var keychainHelper: KeychainHelper
```
adrien-coye, line 300: As they are servicies, and are not subject to change identity, it might be better to keep them as lazy iVars(LazyInjectService).

That way we do not re-do the resolution each time we perform a refresh.

PhilippeWeidmann: I don't want to reference self

valentinperignon (approved):

Ambrdctr (approved):

PR #572, opened 2023-02-14
fix(URLSchemeHandler): Add actor for task management

PhilippeWeidmann:


MailCore/Utils/URLSchemeHandler.swift
```diff
@@ -19,43 +19,86 @@
...
+            var request = URLRequest(url: components!.url!)
+            request.addValue(
+                "Bearer \(AccountManager.instance.currentAccount.token.accessToken)",
+                forHTTPHeaderField: "Authorization"
+            )
+            let task = URLSession.shared.dataTask(with: request) { data, response, error in
+                Task { [weak self] in
+                    guard (error as? URLError)?.code != .cancelled else { return }
+
+                    if let response = response {
```
valentinperignon, line 81:
```suggestion
                    if let response {
```

valentinperignon (approved):

Ambrdctr (approved):

MailCore/Utils/URLSchemeHandler.swift
```diff
@@ -19,43 +19,86 @@
...
+            request.addValue(
+                "Bearer \(AccountManager.instance.currentAccount.token.accessToken)",
+                forHTTPHeaderField: "Authorization"
+            )
+            let task = URLSession.shared.dataTask(with: request) { data, response, error in
+                Task { [weak self] in
+                    guard (error as? URLError)?.code != .cancelled else { return }
+
+                    if let response {
+                        await self?.taskProgressHandler.didReceive(response, for: url)
```
adrien-coye, line 82: I would `guard self` asap first, so we do not optional unwrap each time we access it.

MailCore/Utils/URLSchemeHandler.swift
```diff
@@ -19,43 +19,86 @@
...
-                urlSchemeTask.didReceive(response)
-            }
-            if let data = data {
-                urlSchemeTask.didReceive(data)
-                urlSchemeTask.didFinish()
-            }
-            if let error = error {
-                urlSchemeTask.didFailWithError(error)
+        Task {
+            await taskProgressHandler.cancelDataTask(for: url)
```
adrien-coye, line 67: I may not understand this perfectly, yet if for some reason we "navigate" back and forth you can probably send events multiple times for one single URL.

MailCore/Utils/URLSchemeHandler.swift
```diff
@@ -19,43 +19,86 @@
...
+            let task = URLSession.shared.dataTask(with: request) { data, response, error in
+                Task { [weak self] in
+                    guard (error as? URLError)?.code != .cancelled else { return }
+
+                    if let response {
+                        await self?.taskProgressHandler.didReceive(response, for: url)
+                    }
+                    if let data {
+                        await self?.taskProgressHandler.didReceive(data, for: url)
+                        await self?.taskProgressHandler.didFinish(for: url)
```
adrien-coye, line 86: Within an Actor consistency is guaranteed, yet outside one nothing prevents someone else from calling finish() while you are awaiting didReceive() here to return.

MailCore/Utils/URLSchemeHandler.swift
```diff
@@ -19,43 +19,86 @@
...
+                    }
+                    if let error {
```
adrien-coye, line 88: I would probably guard the error first,  didFailWithError() and returning early if any.

_Then_ I would switch let on the two statements like so: (this is pseudo code)
```swift
switch(data, response) {
case (nil response), (data response):
    await self?.taskProgressHandler.didReceive(response, for: url)
case (data nil), (data response):
    // can we group the two didReceive and didFinish in a dedicated call ?
    await self?.taskProgressHandler.didReceiveAndFinish(data, for: url)
}
```

adrien-coye (commented):
I'm availlable to discuss all my comments.

Also if this continues to be of concern we may catch NSExceptions with a wrapper like this:
https://stackoverflow.com/questions/34956002/how-to-properly-handle-nsfilehandle-exceptions-in-swift-2-0/35003095#35003095

MailCore/Utils/URLSchemeHandler.swift
```diff
@@ -19,43 +19,86 @@
...
+                        await self?.taskProgressHandler.didFailWithError(error, for: url)
```
adrien-coye, line 89: // can we group the two didReceive and didFinish in a dedicated call ?
    await self.taskProgressHandler.didReceiveAndFinish(data, for: url)

MailCore/Utils/URLSchemeHandler.swift
```diff
@@ -19,43 +19,86 @@
...
+
+            var components = URLComponents(url: url, resolvingAgainstBaseURL: true)
+            components?.scheme = "https"
+            var request = URLRequest(url: components!.url!)
+            request.addValue(
+                "Bearer \(AccountManager.instance.currentAccount.token.accessToken)",
+                forHTTPHeaderField: "Authorization"
+            )
+            let task = URLSession.shared.dataTask(with: request) { data, response, error in
+                Task { [weak self] in
```
adrien-coye, line 78: Another idea is to remove this Task and dispatch this block in a serial queue instead. We would need to remove the Actor as well to not await anything.

MailCore/Utils/URLSchemeHandler.swift
adrien-coye: discussed with @PhilippeWeidmann

PR #597, opened 2023-02-22
fix(Folders): Refresh folder list when enter foreground

Ambrdctr:


valentinperignon (approved):

valentinperignon (commented):
Could we do this in the SceneDelegate?

PhilippeWeidmann (approved):

PR #607, opened 2023-03-02
fix: Fetch contacts after create account

PhilippeWeidmann:


MailCore/Cache/AccountManager.swift
```diff
@@ -304,6 +304,10 @@ public class AccountManager: RefreshTokenDelegate {
         setCurrentMailboxForCurrentAccount(mailbox: mainMailbox)
         saveAccounts()
 
+        Task {
+            try await currentContactManager?.fetchContactsAndAddressBooks()
```
valentinperignon, line 308: We must fetch contacts and adress books when we create a new contact. Could you do it?

valentinperignon (approved):

Lnamw (approved):

PR #653, opened 2023-03-23
fix: Update dark mode

valentinperignon:


Mail/Views/Bottom sheets/Actions/ReplyActionsView.swift
```diff
@@ -19,6 +19,7 @@
 import InfomaniakCore
 import InfomaniakCoreUI
 import MailCore
+import MailResources
```
PhilippeWeidmann, line 22: Why do we need it ?

PhilippeWeidmann (dismissed):

PR #665, opened 2023-03-29
Realm sections

PhilippeWeidmann:


PhilippeWeidmann:
Realm doesn't support animation for ObservedSectionedResults

PR #677, opened 2023-04-04
feat(NotificationsHelper): Use real body instead of preview

PhilippeWeidmann:


MailCore/Utils/NotificationsHelper.swift
```diff
@@ -130,4 +131,26 @@ public enum NotificationsHelper {
                             NotificationsHelper.UserInfoKeys.messageUid: message.uid]
         return content
     }
+
+    private static func getCleanBodyFromMessage(_ message: Message) -> String {
```
valentinperignon, line 135:
```suggestion
    private static func getCleanBodyFrom(message: Message) -> String 
    // or
    private static func getCleanBody(from message: Message) -> String 
```

valentinperignon (approved):

Ambrdctr (approved):

PR #721, opened 2023-04-28
feat: Signatures

Ambrdctr:


PhilippeWeidmann:
Signature list is the same for each mailbox. Each mailbox should have it's own signature list.

PhilippeWeidmann (approved):

PR #751, opened 2023-05-17
fix: text/plain with correct size

valentinperignon:


Mail/Views/Thread/SelectableTextView.swift
```diff
@@ -32,11 +36,15 @@ struct SelectableTextView: UIViewRepresentable {
...
         textView.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
 
-        textView.text = text
         return textView
     }
 
     func updateUIView(_ uiView: UITextView, context: Context) {
         uiView.text = text
+
+        DispatchQueue.main.async {
```
adrien-coye, line 45: Do we need to wait at least one run loop after setting the text property or would using @MainActor for the `updateUIView` func would suffice?

adrien-coye (approved):
All good, thanks for the changes.

PR #756, opened 2023-05-24
fix(ComposeMessageView): Only save draft if body modified

Ambrdctr:


adrien-coye (approved):
Discussed in DM, works with signatures too.

PR #758, opened 2023-05-24
Inline images mail slow to display

adrien-coye:
## Abstract 
Performance issues on displaying large emails with inline pictures.
The mainthread was pinned, batching image processing greatly improves the UX in my testing.

## Misc
Mutating the dom of the email is super costly, therefore I used batching to process images and mutate less the content of the email view that is rendered on the main thread.
I went from 300 sec to 50 ish for a mail with a lot of small images, while keeping the UI responsive

### DIFF
--- a/Mail/Views/Thread/MessageView.swift
+++ b/Mail/Views/Thread/MessageView.swift
@@ -54,7 +54,7 @@ extension Sequence {
 // TODO: move to core
 extension Collection {
     /// Returns the element at the specified index if it is within bounds, otherwise nil.
-    subscript (safe index: Index) -> Element? {
+    subscript(safe index: Index) -> Element? {
         return indices.contains(index) ? self[index] : nil
     }
 }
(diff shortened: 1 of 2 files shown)


Mail/Views/Thread/MessageView.swift
```diff
@@ -100,22 +134,52 @@ struct MessageView: View {
...
+                        of: "cid:\(contentId)",
+                        with: "data:\(attachment.mimeType);base64,\(data.base64EncodedString())"
+                    )
+                }
+
+                // Mutate DOM
+                self.insertInlineAttachment(body: body, compactBody: compactBody)
+                
+                // Delay between each chunk processing just enough, so the user feels the UI is responsive.
+                try await Task.sleep(nanoseconds: 4_000_000_000)
```
adrien-coye, line 173: Nice to have, use a debounce instead maybe ?

Mail/Views/Thread/MessageView.swift
```diff
@@ -25,6 +25,40 @@ import RealmSwift
 import Shimmer
 import SwiftUI
 
+// TODO: move to Core
```
adrien-coye, line 28: TODO: move to Core

adrien-coye: Moved to core, but kMail is locked on an old ios-Core, so we keep it here in the meantime.

Mail/Views/Thread/MessageView.swift
```diff
@@ -100,22 +134,52 @@ struct MessageView: View {
...
-
-                presentableBody.body?.value = presentableBody.body?.value?.replacingOccurrences(
-                    of: "cid:\(contentId)",
-                    with: "data:\(attachment.mimeType);base64,\(attachmentData.base64EncodedString())"
-                )
-                presentableBody.compactBody = presentableBody.compactBody?.replacingOccurrences(
-                    of: "cid:\(contentId)",
-                    with: "data:\(attachment.mimeType);base64,\(attachmentData.base64EncodedString())"
-                )
+        Task {
```
adrien-coye, line 137: try detached

Ambrdctr (approved):

PR #759, opened 2023-05-25
Large complex mails are slow to display

adrien-coye:
## Investigation
Preprocessing of the mail is blowing up with large ones.
The preprocessing we do with "swift soup" is blowing up not in a manner I can quickly optimize.

## Proposed solution
Therefore I used an heuristic to disable preprocessing of emails larger than 5meg.
Works like a charm, email is still displayed resonably well, and some other mail clients use this trick.

“Possibly the most common error of a smart engineer is to optimise a thing that should not exist”  - Musk

### DIFF
--- a/MailCore/Utils/MessageBodyUtils.swift
+++ b/MailCore/Utils/MessageBodyUtils.swift
@@ -75,7 +75,6 @@ public enum MessageBodyUtils {
 
     private static func findFirstKnownParentQuoteDescriptor(htmlDocumentWithoutQuote: Document) throws -> String {
         var currentQuoteDescriptor = ""
-        // note: walk the entire dom each time, can be optimized but not the main issue
         for quoteDescriptor in quoteDescriptors {
             let quotedContentElement = try selectElementAndFollowingSiblings(
                 document: htmlDocumentWithoutQuote,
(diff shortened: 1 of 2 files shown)


MailCore/Utils/MessageBodyUtils.swift
```diff
@@ -75,6 +75,7 @@ public enum MessageBodyUtils {
 
     private static func findFirstKnownParentQuoteDescriptor(htmlDocumentWithoutQuote: Document) throws -> String {
         var currentQuoteDescriptor = ""
+        // note: walk the entire dom each time, can be optimized but not the main issue
```
adrien-coye, line 78: todo clean

MailCore/Utils/MessageBodyUtils.swift
```diff
@@ -99,6 +100,7 @@ public enum MessageBodyUtils {
             }
             return try (htmlDocumentWithQuote.outerHtml(), blockquoteElement?.outerHtml())
         } else if !currentQuoteDescriptor.isEmpty {
+            // This is the main hog
```
adrien-coye, line 103: TODO clean

Ambrdctr (approved):

PR #766, opened 2023-06-02
fix: Fetch all new messages after delta

Ambrdctr:


MailCore/Utils/Array+Extension.swift
```diff
@@ -50,3 +50,9 @@ public extension LazyMapSequence {
         return Array(self)
     }
 }
+
+public extension Set {
```
adrien-coye, line 54: move to core ?

adrien-coye: More specificaly, can we move the base type extensionn  to core ?
The specific Realm code ca stay here.

public extension List {
public extension LazyFilterSequence {
public extension LazyMapSequence {

MailCore/Cache/MailboxManager.swift
```diff
@@ -699,16 +703,24 @@ public class MailboxManager: ObservableObject {
...
-            .shortUid?.toString() else { return false }
+        var paginationInfo: PaginationInfo?
+
+        if let offset = realm.objects(Message.self).where({ $0.folderId == folder.id })
+            .sorted(by: {
+                if direction == .following {
+                    return $0.shortUid! > $1.shortUid!
+                }
+                return $0.shortUid! < $1.shortUid!
+            }).first?.shortUid?.toString(), let direction {
```
valentinperignon, line 716:
```suggestion
            }).first?.shortUid?.toString(),
     let direction {
```

MailCore/Utils/Array+Extension.swift
adrien-coye: discussed, will be done in another PR

valentinperignon (approved):

PR #769, opened 2023-06-02
Translate well known notifications

adrien-coye:
Translate only the "new mail" notification + unit tests


PhilippeWeidmann:
This function does pretty much the same as `prepareEmptyNotification`

The reason why the message is not translated is because the notification extension either crashed or was being updated while we received a message.

PR #771, opened 2023-06-05
fix: Current toolbar break when the screen is too small

valentinperignon:
We have 5 items on our bottom bar. With a small screen, the layout is broken, so we now use our own component.


adrien-coye (approved):
discussed in DM, all good

PR #772, opened 2023-06-05
Keyboard select best match on enter

adrien-coye:
- If the email is not in contacts but valid, we try to use it.

### DIFF
--- a/Mail/Components/RecipientField.swift
+++ b/Mail/Components/RecipientField.swift
@@ -71,39 +71,14 @@ struct RecipientField: View {
     }
 
     @MainActor private func submitTextField() {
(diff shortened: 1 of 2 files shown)


Mail/Components/RecipientField.swift
```diff
@@ -108,7 +106,14 @@ struct RecipientField: View {
         }
     }
 
+    private var trimmedInputText: String {
```
valentinperignon, line 109: Maybe we should move this variable above the functions?

adrien-coye: done

valentinperignon (approved):

PR #774, opened 2023-06-06
fix: Selection and multiple selection should have the same vertical padding

valentinperignon:
Currently on iPad the selection background (gray color) and the multiple selection background (pink color) have different vertical padding.
It results in backgrounds with different heights.


adrien-coye (approved):
Looking good 👍

PR #801, opened 2023-06-13
Some messages never finish preprocessing

adrien-coye:
### Abstract

Observing an extra flag, allows us to finish the loading of the view.

### DIFF
--- a/Mail/Views/Thread/MessageBodyView.swift
+++ b/Mail/Views/Thread/MessageBodyView.swift
@@ -16,16 +16,20 @@
  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import InfomaniakCore
 import MailCore
 import MailResources
 import RealmSwift
 import SwiftUI
 
 struct MessageBodyView: View {
+    let taskQueue = TaskQueue()
+
     @State private var textPlainHeight = CGFloat.zero
 
     @StateObject private var model = WebViewModel()
 
+    @Binding var isMessagePreprocessed: Bool
     @Binding var presentableBody: PresentableBody
     var blockRemoteContent: Bool
     @Binding var displayContentBlockedActionView: Bool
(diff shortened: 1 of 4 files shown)


Mail/Views/Thread/MessageView.swift
```diff
@@ -88,12 +87,16 @@ struct MessageView: View {
...
                     MessageBodyView(
+                        isMessagePreprocessed: $isMessagePreprocessed,
                         presentableBody: $presentableBody,
                         blockRemoteContent: isRemoteContentBlocked,
                         displayContentBlockedActionView: $displayContentBlockedActionView,
                         messageUid: message.uid
                     )
                     .padding(.top, 16)
+                    .onChange(of: isMessagePreprocessed) { _ in
+                        // Keep it for event propagation to work
```
adrien-coye, line 98: Without this, observing does not want to propagate

Mail/Views/Thread/MessageBodyView.swift
```diff
@@ -83,19 +84,23 @@ struct MessageBodyView: View {
     }
 
     private func loadBody(blockRemoteContent: Bool) {
-        Task {
+        Task.detached {
```
PhilippeWeidmann, line 87: `Task.detached` (and `MainActor.run`) are not needed since `loadHTMLString` already runs on a background thread

Mail/Views/Thread/MessageView.swift
PhilippeWeidmann: Maybe `.id("\(message.uid) \(isMessagePreprocessed)")`

Mail/Views/Thread/MessageBodyView.swift
adrien-coye: yeah it is overkill here I agree.

adrien-coye: done

PhilippeWeidmann (approved):

PR #811, opened 2023-06-15
feat(mailEditor): Darkmode improvements on pasting rich text content

adrien-coye:


Mail/Helpers/RichTextEditor.swift
```diff
@@ -96,11 +96,29 @@ struct RichTextEditor: UIViewRepresentable {
...
-            if parentBody != content {
-                parent.body = content
+            let parentBody = parent.body
+            Task.detached {
+                var parentBody = parentBody
+                parentBody = parentBody.trimmingCharacters(in: .whitespacesAndNewlines)
+                parentBody = parentBody.replacingOccurrences(of: "\r", with: "")
+                
+                // Remove explicit dark text css and caret for simple darkmode support.
+                parentBody = parentBody.replacingOccurrences(of: "caret-color: rgb(0, 0, 0);", with: "")
```
adrien-coye, line 108: This works, but could be more aesthetic with a regex to handle the ";"

Mail/Helpers/RichTextEditor.swift
```diff
@@ -29,7 +29,7 @@ struct RichTextEditor: UIViewRepresentable {
     typealias UIViewType = MailEditorView
 
     @Binding var model: RichTextEditorModel
-    @Binding var body: String
+    @MainActor @Binding var body: String
```
adrien-coye, line 32: I pin the binding to the mainActor, without it I have a realm crash.

Mail/Helpers/RichTextEditor.swift
```diff
@@ -96,11 +96,29 @@ struct RichTextEditor: UIViewRepresentable {
...
         }
 
+        @MainActor
         func editorContentChanged(_ editor: SQTextEditorView, content: String) {
-            var parentBody = parent.body.trimmingCharacters(in: .whitespacesAndNewlines)
-            parentBody = parentBody.replacingOccurrences(of: "\r", with: "")
-            if parentBody != content {
-                parent.body = content
+            let parentBody = parent.body
+            Task.detached {
```
adrien-coye, line 102: IMO we need the detached here, the larger the content edited / pasted the more we need it.

Mail/Helpers/RichTextEditor.swift
PhilippeWeidmann: The comment above line 108 seems outdated ?

PhilippeWeidmann:
The approach is clever but we need a more versatile solution. The dark mode should be handled not only for rgb 0 0 0 but all dark shade colors. CSS color filter could be a solution but will need time and tweaking.

PR #814, opened 2023-06-16
feat: Update ComposeMessageView

valentinperignon:


Mail/Views/New Message/AutocompletionView.swift
```diff
@@ -18,48 +18,78 @@
...
+                    if !isLastRecipient {
+                        IKDivider()
+                    }
                 }
             }
+        }
+        .onAppear {
+            updateAutocompletion(currentSearch)
+        }
+        .onChange(of: currentSearch, perform: updateAutocompletion)
```
adrien-coye, line 57: To have a cohesive experience with the search, should we debounce this as well ? What do you think ?

This is what we have in the search:
` .debounce(for: .seconds(0.3), scheduler: DispatchQueue.main)`

adrien-coye (approved):
I ran the code, It looks good honestly.

I have just one optional remark regarding debouncing, let me know what you think.
Thanks 👍

Mail/Views/New Message/AutocompletionView.swift
valentinperignon: Seems a good idea, I will try it!

adrien-coye: Perfect, thanks for trying it out.

PhilippeWeidmann (approved):

PR #816, opened 2023-06-16
Catch click in web view

adrien-coye:
PoC for test and review only, does not work given `RichTextEditor` architecture


adrien-coye:
This was for review only, closing now.

PR #817, opened 2023-06-19
refactor way of getting default Signatures

adrien-coye:
### Abstract

We have some signals that we should refactor this. After discussion we chose to refactor the way we store signatures and when we refresh them.

Also contain some bugfix

NOTE: "Change default signature should update in draft" will be done as a separate ticket.

### Changes

- We no longer store the fetch result structure, only a list of Signatures.
- Like discussed, we make sure to refresh the signatures before openning an editor for now.
  - On error, snackbar + dismiss 

Bugfix:
- Extrneous Drafts created when openning an editor.
- Updating a local draft from the server would nilify the signature id, needed to save said draft.
- Fixed empty draft detection and signature insertion if needed, backed by a set of integration tests.


MailCore/Cache/MailboxManager.swift
```diff
@@ -123,26 +122,51 @@ public class MailboxManager: ObservableObject {
...
+            }
+
+            signaturesToUpdate = updatedSignatures.filter { updatedElement in
+                existingSignatures.contains(updatedElement)
+            }
+
+            signaturesToDelete = existingSignatures.filter { existingElement in
+                !updatedSignatures.contains(existingElement)
+            }
+
```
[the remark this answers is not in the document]
adrien-coye: done

Mail/Views/New Message/ComposeMessageBodyView.swift
```diff
@@ -68,8 +71,21 @@ struct ComposeMessageBodyView: View {
...
+        .onChange(of: signatureManager.loadingSignatureState, perform: { state in
+            switch state {
+            case .success:
+                Task {
+                    await setSignature()
+                }
+            case .error(let _):
+                // Unable to get signatures, "An error occurred" and close modal.
+                IKSnackBar.showSnackBar(message: MailError.unknownError.localizedDescription)
+                self.dismiss()
```
PhilippeWeidmann, line 84: self not needed

Mail/Views/New Message/ComposeMessageBodyView.swift
```diff
@@ -68,8 +71,21 @@ struct ComposeMessageBodyView: View {
         }
         .task {
             await prepareReplyForwardBodyAndAttachments()
-            await setSignature()
         }
+        .onChange(of: signatureManager.loadingSignatureState, perform: { state in
```
PhilippeWeidmann, line 75: trailing syntax is prefered

Mail/Views/New Message/ComposeMessageBodyView.swift
```diff
@@ -68,8 +71,21 @@ struct ComposeMessageBodyView: View {
...
+            switch state {
+            case .success:
+                Task {
+                    await setSignature()
+                }
+            case .error(let _):
```
PhilippeWeidmann, line 81: Unused arguments in switch enum can be removed

MailCore/Models/Signature.swift
```diff
@@ -60,9 +56,34 @@ public class Signature: Object, Codable, Identifiable {
...
+            return first
+        }
+
+        // We matched one
+        return defaultSignature
+    }
 }
 
-public class ValidEmail: Object, Decodable {
+public final class ValidEmail: Object, Decodable {
```
PhilippeWeidmann, line 86: This object is not used can you remove it ? (I think it has not been used for quite some time now)

MailCore/Models/Signature.swift
```diff
@@ -60,9 +56,34 @@ public class Signature: Object, Codable, Identifiable {
         // Property 'hash' already exists
         case hashString = "hash"
     }
+
+    override public var hash: Int {
+        id
+    }
+
+    override public func isEqual(_ object: Any?) -> Bool {
```
PhilippeWeidmann, line 64: Why do we need custom `hash` and `isEqual` implementation ?

adrien-coye: The [implementation of the "isEqual" method on the Realm side](https://github.com/realm/realm-swift/blob/354c7ad1827c3f2ec4b51da466c73926984ec245/Realm/RLMObjectBase.mm#LL483C4-L483C4) takes in the realm attached, the detached state and more.

Which is correct, but kind of breaks with our pattern using `.freezeIfNeeded` when I care about something existing or not in a set for a primary key.

Another more verbose way is to filter collections explicitly comparing only on the \id but I'm not sure I like it either.

I guess I can refactor to make sure all the realm objects are both detached in order to _pass_ comparision.

adrien-coye: Confirmed in DM I'm keeping it this way.

adrien-coye:
I still have issues on my Draft folder, please do not merge.

MailCore/Models/Draft.swift
```diff
@@ -258,12 +258,9 @@ public class Draft: Object, Codable, Identifiable {
                      cc: recipientHolder.cc)
     }
 
-    public func setSignature(_ signatureResponse: SignatureResponse) {
-        identityId = "\(signatureResponse.defaultSignatureId)"
-        guard let signature = signatureResponse.default else {
-            return
-        }
-
+    // TODO: Factorise
```
adrien-coye, line 261: Factorise this code part, it may fix the issues I have with my account

adrien-coye: Working on duplicated signatures.

Mail/Views/New Message/Signatures/SignaturesManager.swift
```diff
@@ -0,0 +1,78 @@
...
+        loadRemoteSignatures()
+    }
+
+    /// Load the signatures every time at init, set `doneLoadingDefaultSignature` to true when done
+    private func loadRemoteSignatures() {
+        Task {
+            do {
+                // load all signatures every time
+                try await mailboxManager.refreshAllSignatures()
+                assert(mailboxManager.getStoredSignatures().defaultSignature != nil, "Expecting a default signature")
```
PhilippeWeidmann, line 59: In production we have to dismiss the view in case we fall in the assert condition. Something is missing to trigger loadingSignatureState to error ?

MailCore/Models/Draft.swift
adrien-coye: This is now fixed.

Mail/Views/New Message/Signatures/SignaturesManager.swift
adrien-coye: This asert is checking something that I believe was always true by design, producing a crash in DEBUG.
You are correct I can check it for real in code just to be sure. Will push ASAP.

adrien-coye: done

Mail/Views/New Message/Signatures/SignaturesManager.swift
```diff
@@ -0,0 +1,88 @@
...
+
+import MailCore
+import Sentry
+import SwiftUI
+
+final class SignaturesManager: ObservableObject {
+    /// Local errors for SignaturesManager type
+    private enum ErrorDomain: Error {
+        /// After an update from the server we are still without a default signature
+        case defaultSignatureMissing
```
PhilippeWeidmann, line 27: Declare the error in MailError.defaultSignatureMissing

Mail/Views/Thread List/ThreadListView.swift
```diff
@@ -44,7 +44,8 @@ struct ThreadListView: View {
     @AppStorage(UserDefaults.shared.key(.threadDensity)) private var threadDensity = DefaultPreferences.threadDensity
     @AppStorage(UserDefaults.shared.key(.accentColor)) private var accentColor = DefaultPreferences.accentColor
 
-    @State private var isShowingComposeNewMessageView = false
+    /// Assign a draft will open an edit modal for it.
+    @State private var newDraft: Draft?
```
PhilippeWeidmann, line 48: If variable needs a comment to describe it's function maybe rename it something more meaningful ? eg.
```suggestion
    @State private var newPresentedDraft: Draft?
```
Because views in SwiftUI are function of the state we directly understand that if `newPresentedDraft` is not nil something will be currently presented

adrien-coye: Sure, I like `newPresentedDraft`

PhilippeWeidmann (approved):

PR #844, opened 2023-06-29
feat(MessageView): Improve mail when width changes

valentinperignon:


adrien-coye (commented):
As discussed in DM I think the JS functions could be listed in an enum, beyond that I'm ok with this PR.

adrien-coye (approved):
LGTM 👍

PR #845, opened 2023-06-30
feat: Advertise NSUserActivity

PhilippeWeidmann:
For now we only support navigating to correct mailbox and folder as we don't share thread ids.
This only works if the connected account on the web is the same as on device.

### DIFF
--- a/MailCore/Utils/UserActivityController.swift
+++ b/MailCore/Utils/UserActivityController.swift
@@ -30,13 +30,13 @@ public class UserActivityController {
         folder: Folder?
     ) {
         guard let folder,
-              let mailboxIndex = getMailboxIndexForCustomOrder(mailbox),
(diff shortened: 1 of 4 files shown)


MailCore/Utils/UserActivityController.swift
```diff
@@ -0,0 +1,55 @@
...
+
+public class UserActivityController {
+    var currentActivity: NSUserActivity?
+
+    public init() {}
+
+    public func setCurrentActivity(
+        _ activity: NSUserActivity = NSUserActivity(activityType: NSUserActivityTypeBrowsingWeb),
+        mailbox: Mailbox,
+        folder: Folder?
```
valentinperignon, line 30: Is it ok to have a function that accepts an optional but checks directly that folder is not an optional?

valentinperignon (approved):

PR #846, opened 2023-07-03
fix(Menu): Indicator for unread messages in Folder

Ambrdctr:


MailCore/Models/Message.swift
```diff
@@ -111,6 +114,7 @@ public struct MessagesUids {
     public var deletedUids = [String]()
     public var updated = [MessageFlags]()
     public let cursor: String
+    public var folderUnreadCount: Int? = nil
```
PhilippeWeidmann, line 117: redundant_optional_initialization
```suggestion
    public var folderUnreadCount: Int?
```

PhilippeWeidmann (approved):

PR #857, opened 2023-07-06
fix: Correctly load Draft body

PhilippeWeidmann:
Write complete body in realm only once after complete loading.
Add signature after initial loading.
Only add signature for new messages, reply, forward.


adrien-coye (approved):
I added a note for a nice to have. Also after some time testing I found no issue with this PR.

PR #880, opened 2023-07-14
feat(ContactManager): Try to match email address then name when matching contact

PhilippeWeidmann:
Depends on #879


valentinperignon (approved):
LGTM

PR #886, opened 2023-07-19
fix: Always register notification token and add delay before registering

PhilippeWeidmann:
Because of a master / slave replication issue on the back-end side we cannot register a notification token right after refreshing / creating an API token.


adrien-coye (approved):
LGTM after a DM

PR #892, opened 2023-07-20
refactor: Decouple current account

PhilippeWeidmann:
Ground work stop using AccountManager.currentAccount.
For now all accesses must be done using: `getCurrentAccount()`
In the future if we want to stop using "one current account" eg. show multiple windows with different mailboxes and accounts we will just have to replace `getCurrentAccount()` with `getPreferredAccount()`.

Depends on #890

### DIFF
--- a/Mail/Utils/NavigationState.swift
+++ b/Mail/Utils/NavigationState.swift
@@ -18,6 +18,7 @@
 
 import Combine
 import Foundation
+import InfomaniakCore
 import InfomaniakCoreUI
 import InfomaniakDI
 import MailCore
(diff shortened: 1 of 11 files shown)


MailCore/Cache/AccountManager.swift
```diff
@@ -503,7 +507,7 @@ public class AccountManager: RefreshTokenDelegate, ObservableObject {
     }
 
     public func enableBugTrackerIfAvailable() {
-        if currentAccount.user?.isStaff == true {
+        if let currentAccount, currentAccount.user?.isStaff == true {
```
valentinperignon, line 510:
```suggestion
        if let currentAccount, currentAccount.user.isStaff {
```

Mail/Utils/NavigationState.swift
```diff
@@ -92,28 +93,41 @@ class NavigationState: ObservableObject {
...
         accountManagerObservation = accountManager.objectWillChange.receive(on: RunLoop.main).sink { [weak self] in
-            self?.switchToCurrentMailboxManagerIfPossible()
+            self?.account = AccountManager.instance.getCurrentAccount()
+            self?.rootViewState = NavigationState.getMainViewStateIfPossible()
         }
     }
 
+    static func getMainViewStateIfPossible() -> RootViewState {
+        let accountManager = AccountManager.instance
+        if let currentAccount = accountManager.getCurrentAccount() {
```
valentinperignon, line 110: Could we write it with a guard statement ?

PhilippeWeidmann: FTR: Discussed IRL

valentinperignon (approved):
LGTM

PR #830, opened 2023-06-22
Mail share extension

adrien-coye:
### Abstract 

ShareExtension setup for the kMail project.
NEW: Support the same files as input as kDrive.

### Dependencies
~[Core PR](https://github.com/Infomaniak/ios-core/pull/45)~
~[Core UI PR](https://github.com/Infomaniak/ios-core-ui/pull/8)~
~TODO Core UI + Core PR~
~Bring back latest [Core changes here](https://github.com/Infomaniak/ios-core/pull/53)~
~[use date as default name Core](https://github.com/Infomaniak/ios-core/pull/58)~
~[use date as default name CoreUI](https://github.com/Infomaniak/ios-core-ui/pull/11)~

### Details

Wraps existing SwiftUI code, no way to use pure SwiftUI currently IMO.

Includes also:

- Small refactor of the Tuist setup to help with one extra target.
- AccountManager is now availlable with DI instead of static ref.
- More code split using DI for ease of running all the code in Extension mode.
- Reworked the Attachment ingestion, to prevent errors in extreme cases.
- Abstracted user messaging mechanisms to work seamlessly in extension mode. 
- Abstracted NSItemProvider to File between kDrive and ikMail

PR #897, opened 2023-07-25
feat: Synchronous decoupled token store

PhilippeWeidmann:


PhilippeWeidmann:
Duplication issues will be resolved when merging #830

MailCore/Cache/TokenStore.swift
```diff
@@ -0,0 +1,73 @@
...
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import Foundation
+import InfomaniakCore
+import InfomaniakDI
+
+public class TokenStore {
+    public enum TokenStoreFetchLocation {
+        case cache
```
valentinperignon, line 25: Do we prefer 
```swift
case cache, keychain
```
or
```swift
case cache
case keychain
```
I noticed that I often write only one `case` but have no preference

Mail/Views/Alerts/LogoutConfirmationView.swift
```diff
@@ -43,12 +43,19 @@ struct LogoutConfirmationView: View {
...
         }
-        AccountManager.instance.removeTokenAndAccount(token: account.token)
-        if let nextAccount = AccountManager.instance.accounts.first {
-            AccountManager.instance.switchAccount(newAccount: nextAccount)
-        }
-        AccountManager.instance.saveAccounts()
+    }
+}
+
+extension ApiToken: Identifiable {
```
valentinperignon, line 56: Why ApiToken needs to be Identifiable here?

Mail/Views/Switch User/AccountView.swift
```diff
@@ -27,13 +27,15 @@ import SwiftUI
 
 class AccountViewDelegate: DeleteAccountDelegate {
     @MainActor func didCompleteDeleteAccount() {
-        guard let account = AccountManager.instance.getCurrentAccount() else { return }
-        AccountManager.instance.removeTokenAndAccount(token: account.token)
-        if let nextAccount = AccountManager.instance.accounts.first {
-            AccountManager.instance.switchAccount(newAccount: nextAccount)
-            IKSnackBar.showSnackBar(message: "Account deleted")
+        Task {
```
valentinperignon, line 30: Can you explain me why we need to create a task

valentinperignon (commented):
👍🏻

Mail/Views/Switch User/AccountView.swift
PhilippeWeidmann: We don't, at first TokenStore was an Actor and we had to await every call but I couldn't do this unless a lot of refactor was done so I stayed with our thread safe (but blocking) collections

Mail/Views/Alerts/LogoutConfirmationView.swift
PhilippeWeidmann: It needs to be identifiable to be presented as a an item in the sheet before. I moved this extension in the view that this presenting the logout confirmation

MailCore/Cache/TokenStore.swift
PhilippeWeidmann: I do not have a preference either but I would tend to say that multiple line cases improve readability and are easier to refactor if we add more cases in the future.

valentinperignon (approved):

PR #899, opened 2023-07-25
feat: Update attachments icons

valentinperignon:


MailCore/Models/Attachment.swift
```diff
@@ -53,22 +53,43 @@ public class Attachment: /* Hashable, */ EmbeddedObject, Codable, Identifiable {
 
     public var icon: MailResourcesImages {
         guard let uti else { return MailResourcesAsset.unknownFile }
+
+        if uti.conforms(to: .pdf) {
```
adrien-coye, line 57: Eventualy (not right now), it would be nice to have this abstracted in {core/ui} so we could share this between drive and mail

MailCore/Utils/UTType+Extension.swift
```diff
@@ -0,0 +1,30 @@
...
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import Foundation
+import UniformTypeIdentifiers
+
+public extension UTType {
```
adrien-coye, line 22: There will be soon an UTI in core that can be extended and used by platform apps. (after share ext is meged)

MailCore/Models/Attachment.swift
```diff
@@ -53,22 +53,43 @@ public class Attachment: /* Hashable, */ EmbeddedObject, Codable, Identifiable {
...
+            return MailResourcesAsset.pdfFile
+        }
+        if uti.conforms(to: .calendarEvent) || uti.conforms(to: .ics) {
```
adrien-coye, line 60: I think I prefer `if {} else if {} else {}` to `if {} if {} return`

If we miss a return statement, the compiler should warn us.

Here is what I did for the [share ext file types](https://github.com/Infomaniak/ios-core/blob/master/Sources/InfomaniakCore/ItemProviderRepresentation/NSItemProvider%2BDetect.swift) Let me know what you think.

adrien-coye (commented):
I did a first review

MailCore/Utils/UTType+Extension.swift
adrien-coye: discussed with @valentinperignon , no need for ikMail. Will check later if we need it for kDrive.

adrien-coye (approved):
Discussed in DM,  LGTM

PR #906, opened 2023-07-28
chore: Swift format pre commit

PhilippeWeidmann:
WIP: Discussing internally the best way to enforce the use of this tool


.swiftformat-bin/swiftformat
adrien-coye, line 1: Can we import swiftformat as a dependecy ? So we can run a version up to date easily, and control the version with tuist maybe, or would this be too complex ?

.swiftformat
```diff
@@ -0,0 +1,85 @@
+--acronyms ID,URL,UUID
```
adrien-coye, line 1: I'm on the fence about putting this in a dedicated module, so it is shared between all apps and can be updated in one place. What to you reckon ?

.swiftformat-bin/swiftformat
adrien-coye: As this tool is availlable [with a GH release](https://github.com/apple/swift-format/releases), maybe [there is an API](https://docs.github.com/en/free-pro-team@latest/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release) that we can hook into.

The git project only contain sources sadly.

adrien-coye (commented):
Discussed in DM, ok to use git hooks. 
Also, I have a couple of remarks regarding the way we could import swift format.

.swiftformat-bin/swiftformat
PhilippeWeidmann: I looked into it and this will be hard to do as tuist is meant only to generate Xcode projects

.swiftformat
PhilippeWeidmann: I'm open to the idea. I tried to do it with an SPM that we could import as a dependency but the path of an SPM is hard to predict.
Maybe as git submodule ? But until now we only have had SPM so it could be tidious to always have to fetch both.

PhilippeWeidmann:
Will be handled using an internal setup tool

PR #910, opened 2023-08-04
fix(BlockUser): Show snackbar after call

Ambrdctr:


PhilippeWeidmann:
Not relevant since #924

PR #915, opened 2023-08-09
Fix archive UI glitch

adrien-coye:
Fix by putting `refreshFolder(from messages)` function into an actor for locking.

### DIFF
--- a/MailCore/Cache/MailboxManager.swift
+++ b/MailCore/Cache/MailboxManager.swift
@@ -245,24 +245,22 @@ public final class MailboxManager: ObservableObject {
         return folder
     }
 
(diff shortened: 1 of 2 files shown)


MailCore/Cache/MailboxManager.swift
```diff
@@ -245,24 +245,22 @@ public final class MailboxManager: ObservableObject {
         return folder
     }
 
+    // MARK: RefreshActor
```
PhilippeWeidmann, line 248:
```suggestion
    // MARK: - RefreshActor
```

PhilippeWeidmann (approved):

PR #917, opened 2023-08-10
feat: Single message mode setting added

Ambrdctr:


MailCore/Cache/AccountManager.swift
```diff
@@ -489,4 +489,10 @@ public final class AccountManager: RefreshTokenDelegate, ObservableObject {
             bugTracker.stopActivatingOnScreenshot()
         }
     }
+    
+    public func updateConversationSettings() {
+        for mailboxManager in mailboxManagers.values {
```
PhilippeWeidmann, line 494: This will only work for already instantiated mailboxmanagers. You have to loop and create each mailboxManager an example can be seen here -> NotificationsHelper.swift:71

Mail/Views/ThreadListManagerView.swift
```diff
@@ -34,6 +36,7 @@ struct ThreadListManagerView: View {
                     SearchView(mailboxManager: mailboxManager, folder: selectedFolder)
                 } else {
                     ThreadListView(mailboxManager: mailboxManager, folder: selectedFolder, isCompact: isCompactWindow)
+                        .id(threadMode)
```
PhilippeWeidmann, line 39: Check if reusing ThreadListView.swift:184 would also work ?

PhilippeWeidmann (approved):

PR #924, opened 2023-08-14
refactor: ActionsManager

PhilippeWeidmann:


adrien-coye (commented):
I read the PR to get an understanding on the implications for my branches.

Mail/Views/Bottom sheets/Actions/ContactActionsView.swift
```diff
@@ -57,13 +57,14 @@ struct ContactActionsView: View {
                     IKDivider()
                 }
 
-                ActionView(action: action) {
+                // FIXME: target for now is only message, not contact
```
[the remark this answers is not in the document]
PhilippeWeidmann: Yup I wouldn't mind having two separated views but I also want to reuse this component as much as possible so I still need to figure out which View code will be reused

MailCore/Cache/Actions/ActionsManager.swift
```diff
@@ -0,0 +1,245 @@
...
+import Foundation
+import InfomaniakCoreUI
+import InfomaniakDI
+import MailResources
+import SwiftUI
+
+extension [Message]: Identifiable {
+    public var id: Int {
+        // Calculate a unique identifier by XORing hash values of messages
+        return reduce(1) { $0.hashValue ^ $1.hashValue }
```
adrien-coye, line 28: I double checked, the order of XOR operations does not affect the final result when you're XORing multiple values together. So this works well actualy, as I think we care about having a stable ID for a `Set` (regardless of ordering) of messages which this does.

MailCore/Cache/Actions/ActionsManager.swift
```diff
@@ -0,0 +1,245 @@
...
+
+    public init(mailboxManager: MailboxManager, navigationState: NavigationState?) {
+        self.mailboxManager = mailboxManager
+        self.navigationState = navigationState
+    }
+
+    public func performAction(target messages: [Message], action: Action, origin: ActionOrigin) async throws {
+        let messagesWithDuplicates = messages.addingDuplicates()
+
+        switch action {
```
adrien-coye, line 78: If we hit the max cyclomatic complexity, we probably should have a function for every action (some already have), or something like that.

adrien-coye (commented):
I did a review an ran the code localy. Looks good.

MailCore/Cache/Actions/ActionsManager.swift
adrien-coye: It looks like we use it for identity of swiftUI views, therefore it might be a bug for `[MessageA, MessageB]` and `[MessageB, MessageA]` to have the same identity 🤔

adrien-coye: tl;dr you can resolve this @PhilippeWeidmann. We only need to keep in mind that if we used a `[Message]` directly in a swiftUI List it will probably not refresh on order change.

MailCore/Cache/Actions/Action.swift
```diff
@@ -0,0 +1,90 @@
...
+
+import Foundation
+import MailResources
+import SwiftUI
+
+public struct Action: Identifiable, Hashable, Equatable {
+    public let id: String
+    public let title: String
+    public let shortTitle: String?
+    public let iconName: String
```
valentinperignon, line 27: Why do we need `iconName` and `tintColorName` ?

MailCore/Cache/Actions/ActionsManager.swift
PhilippeWeidmann: Yup totally agree

Mail/Views/Thread/MoveEmailView.swift
```diff
@@ -87,30 +78,31 @@ struct MoveEmailView: View {
...
-                }
-                dismissModal()
+                move(to: newFolder)
             })
         }
     }
 
+    private func move(to folder: Folder) {
+        Task {
+            // TODO: Also use Actions manager
```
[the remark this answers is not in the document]
PhilippeWeidmann: I think that I missed it, i think a snack bar is missing

MailCore/Cache/Actions/Action.swift
PhilippeWeidmann: It was easier to do it like that to make Action automatically conform to hashable etc. And Image is not hashable

Mail/Views/Thread/MoveEmailView.swift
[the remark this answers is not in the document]
PhilippeWeidmann: Done in latest commits

MailCore/Cache/Actions/ActionsManager.swift
PhilippeWeidmann: Also refactored a bit in latest commits

valentinperignon (approved):
LGTM

PR #926, opened 2023-08-14
Use reportJunk everywhere instead of Spam

adrien-coye:


adrien-coye:
handeled in another ticket.

PR #928, opened 2023-08-15
Create account view

Lnamw:


Mail/Views/Switch User/CreateAccountView.swift
```diff
@@ -0,0 +1,90 @@
...
+import SwiftUI
+
+struct CreateAccountView: View {
+    @AppStorage(UserDefaults.shared.key(.accentColor)) private var accentColor = DefaultPreferences.accentColor
+
+    @State private var isPresentingCreateAccount = false
+    @StateObject private var loginHandler = LoginHandler()
+
+    var body: some View {
+        VStack(alignment: .center, spacing: 0) {
```
Ambrdctr, line 32: Spacing of 24 ?

valentinperignon, line 32:
```suggestion
        VStack(spacing: 0) {
```
The alignment is unnecessary as .center is the default value

Mail/Views/Switch User/CreateAccountView.swift
```diff
@@ -0,0 +1,90 @@
...
+                    .textStyle(.labelMediumAccent)
+                    .frame(maxWidth: .infinity, maxHeight: .infinity)
+                    .padding()
+                    .background(accentColor.secondary.swiftUIColor)
+                    .clipShape(Capsule())
+                    .multilineTextAlignment(.center)
+                Text(MailResourcesStrings.Localizable.newAccountStorageDrive)
+                    .textStyle(.labelMediumAccent)
+                    .frame(maxWidth: .infinity, maxHeight: .infinity)
+                    .padding()
```
valentinperignon, line 55: Shouldn't we assign a value to the padding?

Mail/Views/Switch User/CreateAccountView.swift
Lnamw: As it's not 24 everywhere, I think it's better to not specify it here

PR #929, opened 2023-08-15
fix: Present UIActivityViewController using SwiftUI

PhilippeWeidmann:
Because of a bad merge, RootViewManageable factory was missing and this would cause a crash. However RootViewManageable is not needed anymore as we should always try to present the view from the current scene.

I removed the RootViewManageable protocol and presented the view using a  UIViewControllerRepresentable


adrien-coye (approved):
LGTM

PR #931, opened 2023-08-16
Handling invalid signature, use default one.

adrien-coye:


MailCore/API/MailApiError.swift
```diff
@@ -19,13 +19,30 @@
...
 
+/// Static definitions of Mail API error codes
+enum MailApiErrorCode {
+    /// The email is not found
+    static let mailMessageNotFound = "mail__message_not_found"
+
+    /// Invalid credentials
+    static let invalidCredentials = "invalid_credentials"
+
+    /// The server does not know bout the identity used in the request
```
adrien-coye, line 30:
```suggestion
    /// The server does not know about the identity used in the request
```

MailCore/Cache/MailboxManager/MailboxManager+Draft.swift
```diff
@@ -72,6 +77,11 @@ public extension MailboxManager {
                 }
             }
         } catch let error as MailApiError {
```
PhilippeWeidmann, line 79: Maybe we can even put it in an additional catch close ?

MailCore/Cache/DraftManager.swift
```diff
@@ -92,13 +92,47 @@ public final class DraftManager {
...
+        }
+
+        var updatedDraft: Draft?
+        let realm = mailboxManager.getRealm()
+        try? realm.write {
+            guard let liveDraft = realm.object(ofType: Draft.self, forPrimaryKey: draft.localUUID) else {
+                return
+            }
+            liveDraft.identityId = "\(defaultSignature.id)"
+
```
PhilippeWeidmann, line 127: trailing_whitespace

PhilippeWeidmann:
Apart from the comments the rest looks good to me.

Note that we still need to stop fetching signatures when ComposeView is presented and do it at app foreground appear but this can be done in an other PR.

MailCore/Cache/MailboxManager/MailboxManager+Draft.swift
adrien-coye: Sure I can do a `catch where`. I had in mind to catch all `MailApiError` in the same scope.

adrien-coye: As our errors are not currently `enums` but `objects`, instances of `MailApiError` here.

The syntax is less aestetic, but this does __not__ seems to work 
`catch let error as MailApiError where error == MailApiError.apiIdentityNotFound {}`

I can make it work with this ``catch MailApiError.apiIdentityNotFound {}`__if__ I rewrite all MailApiError with an enum.

PhilippeWeidmann: This needs too much refactor let's keep it this way then

PhilippeWeidmann (approved):

PR #932, opened 2023-08-16
fix: Sort mailbox correctly

Ambrdctr:


adrien-coye (dismissed):
I have not ran the code but it looks good to me.

PhilippeWeidmann (changes requested):
I would place this code in Mailbox.swift like this:

```Swift
extension [Mailbox] {
    func webmailSorted() -> [Mailbox] {
        return sorted {
            if $0.isPrimary {
                return true
            } else if $1.isPrimary {
                return false
            }

            return $0.email < $1.email
        }
    }
}
```
So that it can be reused in UserActivityController.swift

PhilippeWeidmann (approved):

PR #953, opened 2023-08-29
fix(realm_migration): soft migration for Mailbox.aliases 

adrien-coye:
… in order to keep notifications working without openning the app after an update.

- Force REALM migration to take place early in extension mode (notification + share)
- Fix an issue with share extension not loading the correct realm context in edition of a draft.
- Harden contact update with the use of `ApplicationBackgroundTaskTracker` from core-ui

### DIFF
--- a/MailNotificationServiceExtension/NotificationService.swift
+++ b/MailNotificationServiceExtension/NotificationService.swift
@@ -30,6 +30,7 @@ final class NotificationService: UNNotificationServiceExtension {
     private let dependencyInjectionHook = EarlyDIHook()
 
     @LazyInjectService private var accountManager: AccountManager
+    @LazyInjectService private var mailboxInfosManager: MailboxInfosManager
 
     var contentHandler: ((UNNotificationContent) -> Void)?
     var bestAttemptContent: UNMutableNotificationContent?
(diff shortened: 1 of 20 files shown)


adrien-coye:
Duplication by design (independent) between the DI setup of the Extension and the App. We can ignore it safely.

MailNotificationServiceExtension/NotificationService.swift
```diff
@@ -30,13 +30,15 @@ final class NotificationService: UNNotificationServiceExtension {
...
     @LazyInjectService private var accountManager: AccountManager
+    @LazyInjectService private var mailboxInfosManager: MailboxInfosManager
 
     var contentHandler: ((UNNotificationContent) -> Void)?
     var bestAttemptContent: UNMutableNotificationContent?
 
     override init() {
         super.init()
         Logging.initLogging()
+        ModelMigrator().migrateRealmIfNeeded()
```
PhilippeWeidmann, line 41: I didn't check but please test that no call to realm is made before this call else it will be useless. Because EarlyDIHook is called before init and maybe accountManager can be called before ?

adrien-coye: I'll check, the idea was exactly that, to do it early before we handle notifications (in realm) … but with DI in place.

valentinperignon (approved):

PR #958, opened 2023-09-01
Revert "feat: Create account view"

adrien-coye:
Reverts Infomaniak/ios-kMail#928


adrien-coye:
I tested the rollback on iPhone8 test. Nothing to report.

PR #987, opened 2023-09-19
fix: Disable tracing headers introduced with latest Sentry versions

PhilippeWeidmann:
As they are not currently used with by our back-end we do not need to pollute client headers.


adrien-coye (commented):
LGTM 👍

adrien-coye (approved):
LGTM 👍

PR #994, opened 2023-09-25
feat: Basic ThreadListCellContextMenu for macOS

PhilippeWeidmann:


PhilippeWeidmann:
It should work in an extension but we only use it in the ThreadList which is not shown in the extension

adrien-coye (approved):
all good.

PR #998, opened 2023-09-27
fix: Handle message not found

PhilippeWeidmann:


MailCore/Utils/SentryDebug.swift
```diff
@@ -134,4 +134,51 @@ public enum SentryDebug {
             )
         }
     }
+
+    static func listIncoherentMessageUpdate(messages: [Message], actualSeen: Bool) {
+        Task {
+            for message in messages {
+                if let liveMessage = message.thaw(),
```
valentinperignon, line 141: Can we use a guard statement here to avoid the shift of the code?

Ambrdctr (approved):

PR #1004, opened 2023-09-29
feat(AI Writer): Feature flag

valentinperignon:


Mail/Helpers/AppAssembly.swift
```diff
@@ -61,6 +61,9 @@ enum ApplicationAssembly {
             Factory(type: InfomaniakNotifications.self) { _, _ in
                 InfomaniakNotifications(appGroup: AccountManager.appGroup)
             },
+            Factory(type: FeatureFlagsManageable.self) { _, _ in
```
adrien-coye, line 64: Now we have a separate assembly for the NotificationServiceExtension and the Share extension.
We should add the factory there too.

MailCore/Utils/Constants.swift
```diff
@@ -71,6 +71,8 @@ public enum Constants {
         return "const MESSAGE_SELECTOR = \"#\(divWrapperId)\"; \(mungeScript)"
     }()
 
+    public static let defaultFeatureFlags: [FeatureFlag] = []
```
adrien-coye, line 74: I'd add a comment here as it is important I think.

MailCore/Utils/FeatureFlagsManageable.swift
```diff
@@ -0,0 +1,31 @@
...
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import Foundation
+import SwiftUI
+
+public protocol FeatureFlagsManageable {
```
adrien-coye, line 22: We should probably add a minimum of documentation on public APIs even if it looks self-evident ?

MailCore/Utils/FeatureFlagsManageable.swift
```diff
@@ -0,0 +1,31 @@
...
+    typealias UserId = Int
+    typealias AppFeatureFlags = [UserId: [FeatureFlag]]
```
adrien-coye, line 24: A quickwin could be using a [SendableDictionary.swift](https://github.com/Infomaniak/ios-core/blob/master/Sources/InfomaniakCore/Asynchronous/SendableDictionary.swift) it would make the manager thread safe for free.

probably looks like this `SendableDictionary<[UserId: [FeatureFlag]]>`

adrien-coye (commented):
Looking good 👍

Mail/Helpers/AppAssembly.swift
adrien-coye: The two classes used by the App and both share extension are `ApplicationAssembly` and `NotificationServiceAssembly`

MailCore/Utils/FeatureFlagsManageable.swift
adrien-coye: (I think making an actor will be more troublesome, but better to do if you can make it work)

adrien-coye (approved):
All good 👍

PR #1009, opened 2023-10-02
fix: Use floatingPanel from coreUI

Ambrdctr:


valentinperignon (commented):
We should merge "floating-panel" and we still need swift-backport in the project

PR #1019, opened 2023-10-04
fix: NavBar with long title

Lnamw:


valentinperignon (commented):
I have a few questions about how the toolbar works and the proposed fix.. We'll talk about it monday! 😊

valentinperignon (approved):

PR #1044, opened 2023-10-12
fix(Draft): Save drafts when a modification or content is added to the signature of a Draft

adrien-coye:


MailCore/Models/Draft.swift
```diff
@@ -336,13 +344,28 @@ public extension Draft {
         guard let signatureNode = try? document.getElementsByClass(Constants.signatureWrapperIdentifier).first() else {
             return !document.hasText()
         }
+
+        // We check if the signature was changed. If it was edited, we do want to save the draft.
+        // User could have taped inside the signature div and started a Dostoevsky class novel.
+        let signatureNodeText: String? = try? signatureNode.text()
+        guard let rawSignature,
```
valentinperignon, line 351: Instead of storing the raw signature in Realm, maybe we can compare the current signature (`signatureNodeText`) and the signature stored in realm (thanks to `identityId`)?

adrien-coye: Yeah but then what should we do if the signature is removed from the server (and also the phone), or just changed in mid air when we do perform the check ?

adrien-coye:
Discussed with @valentinperignon, the issue goes beyond the empty signatures (and this PR solves the problem for arbitrary signatures). Will check with @PhilippeWeidmann when possible.

MailCore/Cache/DraftContentManager.swift
```diff
@@ -167,6 +167,7 @@ public class DraftContentManager: ObservableObject {
             if liveIncompleteDraft.identityId == nil || liveIncompleteDraft.identityId?.isEmpty == true {
                 liveIncompleteDraft.identityId = "\(signature.id)"
                 if shouldAddSignatureText {
+                    liveIncompleteDraft.rawSignature = signature.content
```
adrien-coye, line 170: Check that we update as well when we change the signature

MailCore/Models/Draft.swift
```diff
@@ -336,13 +344,27 @@ public extension Draft {
         guard let signatureNode = try? document.getElementsByClass(Constants.signatureWrapperIdentifier).first() else {
             return !document.hasText()
         }
+
+        // We check if the signature was changed, the user might also have written within the signature div without knowing.
+        let signatureNodeText: String? = try? signatureNode.text()
```
valentinperignon, line 349:
```suggestion
        let signatureNodeText = try? signatureNode.text()
```

valentinperignon (commented):
Can you update `rawSignature` when we `DraftContentManager.updateSignature` ?

adrien-coye:
I did the update raw signatue on `DraftContentManager.updateSignature`. Just did some manual testing, good to merge.

valentinperignon (approved):

PR #1052, opened 2023-10-13
fix: Prevent SplitView from reloading twice when we switch mailbox.

adrien-coye:
Solved the issue by changing the scope of SplitViewManager.


adrien-coye:
checked with @PhilippeWeidmann . This is a hack, the folder handling should be refactored in order to fix this cleanly.

adrien-coye:
checked with @PhilippeWeidmann . This is a hack, the folder handling should be refactored in order to fix this cleanly.

PR #1054, opened 2023-10-16
fix: Sync UI fixes

PhilippeWeidmann:


valentinperignon (changes requested):
Need to update matomo events

valentinperignon (approved):

PR #1058, opened 2023-10-17
feat: Show update available 

Ambrdctr:


Mail/MailApp.swift
```diff
@@ -67,6 +67,8 @@ struct MailApp: App {
                 .onChange(of: scenePhase) { newScenePhase in
                     switch newScenePhase {
                     case .active:
+                        @LazyInjectService var appLaunchCounter: AppLaunchCounter
```
valentinperignon, line 70:
```suggestion
                        @InjectService var appLaunchCounter: AppLaunchCounter
```

Mail/Views/SplitView.swift
```diff
@@ -101,12 +109,33 @@ struct SplitView: View {
...
             }
         }
+        .floatingPanel(isPresented: $isShowingUpdateAvailable) {
+            DiscoveryView(item: .updateDiscovery) { /* Empty on purpose */ } completionHandler: { update in
+                guard update else { return }
+                let url: URLConstants = Bundle.main.isRunningInTestFlight ? .testFlight : .appStore
+                openURL(url.url)
+            }
+        }
+        .floatingPanel(isPresented: $isShowingSyncDiscovery, content: {
```
valentinperignon, line 119: Can we do something like this?
```suggestion
        .floatingPanel(isPresented: $isShowingSyncDiscovery) {
```

Mail/MailApp.swift
adrien-coye, line 70: As a rule of thumb, it is better to use the DI this way:

- `LazyInjectService` as an ivar (like the account manager) _if possible_ so the resolution is done once.
If not possible to use an ivar, fallback with:
- `InjectService` used directly inside the scope of a function, the lazy will not be usefull here.

Here I think you can move this `@LazyInjectService var appLaunchCounter: AppLaunchCounter` just under `@LazyInjectService private var accountManager: AccountManager`

Mail/Views/SplitView.swift
```diff
@@ -228,4 +257,13 @@ struct SplitView: View {
...
         }
     }
+
+    private func showSync() -> Bool {
+        guard UserDefaults.shared.shouldPresentSyncDiscovery,
+              !appLaunchCounter.isFirstLaunch else {
+            return false
+        }
+
+        return appLaunchCounter.value > 5
```
adrien-coye, line 267: maybe move the `5` to Constants.swift?

Mail/Views/SplitView.swift
```diff
@@ -101,12 +109,33 @@ struct SplitView: View {
...
+            SyncProfileNavigationView()
+        }
         .sheet(item: $navigationState.editedDraft) { editedDraft in
             ComposeMessageView(editedDraft: editedDraft, mailboxManager: mailboxManager)
         }
         .onChange(of: scenePhase) { newScenePhase in
             guard newScenePhase == .active else { return }
             Task {
+                // We don't want to show both DiscoveryView at the same time
+                isShowingUpdateAvailable = try await VersionChecker.standard.showUpdateVersion()
```
valentinperignon, line 137: Can you move the call `try await VersionChecker.standard.showUpdateVersion()` below `async let _ = try? mailboxManager.refreshAllSignatures()` so it won't block the execution?

Mail/Helpers/AppAssembly.swift
```diff
@@ -121,6 +121,9 @@ enum ApplicationAssembly {
             },
             Factory(type: ConfigWebServer.self) { _, _ in
                 ConfigWebServer()
+            },
+            Factory(type: AppLaunchCounter.self) { _, _ in
```
adrien-coye, line 125: There is another Assembly for the share extension. 
I think the Extension will work without knowing about the `AppLaunchCounter` since it is used only in the `SplitView` and `MailApp`.

We should verify the share extension still work to be sure.

adrien-coye (commented):
I did a review related to the use of DI.

Mail/Helpers/AppAssembly.swift
Ambrdctr: Not working.

valentinperignon (approved):

PR #1069, opened 2023-10-18
fix: Do not remove token when displaying DeleteAccountView

valentinperignon:


adrien-coye (approved):
Thx for the i18n fix.

PR #1076, opened 2023-10-23
fix: Clear Realm files on incompatibility for macOS

adrien-coye:
#### Note

Also factorised multiple `.getRealm()` implementations behind a `RealmAccessible` protocol.


MailCore/Cache/MailboxManager/MailboxManager+DB.swift
```diff
@@ -0,0 +1,42 @@
...
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import Foundation
+
+public extension MailboxManager {
```
valentinperignon, line 21: As `RealmAccessible` is related to DB, maybe we can conform to this protocol in this file?

adrien-coye: Yes I actualy created this file to put it here to begin with. There is an issue where if I put the conformance in an extension here, code outside core fails to access it. (you cannot declare extesions as public) so this is what I came with.

valentinperignon:
LGTM

valentinperignon (approved):

PR #1084, opened 2023-10-26
feat(RoleFoldersListView): Can collapse RoleFolders

Lnamw:


valentinperignon (commented):
Dear Elena,

It was a pleasure to review your PR 1084. I’m taking up my pen today to give you some feedback.

This following code is duplicated with `UserFoldersListView`:
```swift
private let folders: [NestableFolder]
private let hasSubFolders: Bool

init(folders: [NestableFolder]) {
 self.folders = folders
 hasSubFolders = folders.contains { !$0.children.isEmpty }
}
```

I believe we can refactor the code.
RoleFoldersListView might become FoldersListView because it is not specific to role folders. And `UserFoldersListView` could use this component.

If you have any question, ask Adrien ;)

valentinperignon (approved):

PR #1086, opened 2023-10-30
ci: Update check commit message script

PhilippeWeidmann:


adrien-coye (approved):
Perfect, we can use any char within  the `()` 👍

PR #1089, opened 2023-10-31
feat(Mac+iPad): Shortcut added

Ambrdctr:


Mail/Utils/ShortcutModifier.swift
```diff
@@ -0,0 +1,131 @@
...
+                    .keyboardShortcut(.delete, modifiers: [])
+
+                Button(MailResourcesStrings.Localizable.actionReply, action: shortcutReply)
+                    .keyboardShortcut("r")
+
+                Button(MailResourcesStrings.Localizable.buttonNewMessage, action: shortcutNewMessage)
+                    .keyboardShortcut("n")
+
+                Button(MailResourcesStrings.Localizable.shortcutRefreshAction, action: shortcutRefresh)
+                    .keyboardShortcut("n", modifiers: [.shift, .command])
```
PhilippeWeidmann, line 49: Would be more logical to use shit + ⌘ + R for refresh ?

Ambrdctr: shift + ⌘ + N is the one used by Apple Mail app. So i used the same.

Mail/Utils/ShortcutModifier.swift
```diff
@@ -0,0 +1,131 @@
...
+    func body(content: Content) -> some View {
+        ZStack {
+            VStack {
+                Button(MailResourcesStrings.Localizable.actionDelete, action: shortcutDelete)
+                    .keyboardShortcut(.delete, modifiers: [])
+
+                Button(MailResourcesStrings.Localizable.actionReply, action: shortcutReply)
+                    .keyboardShortcut("r")
+
+                Button(MailResourcesStrings.Localizable.buttonNewMessage, action: shortcutNewMessage)
```
PhilippeWeidmann, line 45: I understand that you put it here to have all the shortcuts in the same place but I think it would be more relevant to have this one attached to the compose button. What do you think ?

Ambrdctr: Good idea for this one.
I'll try to add it to the real button

Mail/Utils/ShortcutModifier.swift
PhilippeWeidmann: Oh didn't know about this one, okay then 👍

PhilippeWeidmann (approved):

PR #1090, opened 2023-10-31
fix: Feature flags are now per mailbox

PhilippeWeidmann:
~Waiting for confirmation on `email` instead of `mailboxId`~


adrien-coye (approved):
All good 👍 

As a side note, in the future we should think of throwing errors instead of `guard else { return }` when missing something as critical as a MailboxManager.

PR #1096, opened 2023-11-02
fix(MessageView): Use webview all the time

Ambrdctr:


Mail/Views/Thread/MessageBodyView.swift
```diff
@@ -85,14 +75,31 @@ struct MessageBodyView: View {
...
             )
 
             displayContentBlockedActionView = (loadResult == .remoteContentBlocked)
         }
     }
+
+    private func createHtmlForPlainText(text: String?) -> String {
+        do {
+            guard let text else { return "" }
+            let doc = try SwiftSoup.parse("")
```
valentinperignon, line 96: To create an empty document, you should be able to do something like `Document.createShell`

Mail/Views/Thread/MessageBodyView.swift
```diff
@@ -85,14 +75,31 @@ struct MessageBodyView: View {
...
+            try doc.body()?.appendElement("pre").text(text).attr("style", "word-wrap: break-word; white-space: pre-wrap;")
+            return try doc.html()
```
valentinperignon, line 98: Maybe use outerHtml() ?

Mail/Views/Thread/MessageBodyView.swift
```diff
@@ -85,14 +75,31 @@ struct MessageBodyView: View {
...
                 blockRemoteContent: blockRemoteContent
             )
 
             displayContentBlockedActionView = (loadResult == .remoteContentBlocked)
         }
     }
+
+    private func createHTMLForPlainText(text: String?) -> String {
+        do {
+            guard let text else { return "" }
```
valentinperignon, line 95: Maybe you can move `guard` before the block do?
So the catch would be `return text`

Mail/Views/Thread/MessageBodyView.swift
```diff
@@ -85,14 +75,31 @@ struct MessageBodyView: View {
...
-                value: model.showBlockQuote ? presentableBody.body?.value : presentableBody.compactBody,
+                value: messageBody,
                 blockRemoteContent: blockRemoteContent
             )
 
             displayContentBlockedActionView = (loadResult == .remoteContentBlocked)
         }
     }
+
+    private func createHTMLForPlainText(text: String?) -> String {
```
valentinperignon, line 93: Couldn't we do the processing from the webview model?
I don't have a strong opinion, what do you think?

Mail/Views/Thread/MessageBodyView.swift
```diff
@@ -85,8 +75,14 @@ struct MessageBodyView: View {
 
     private func loadBody(blockRemoteContent: Bool) {
         Task {
+            var messageBody = model.showBlockQuote ? presentableBody.body?.value : presentableBody.compactBody
```
valentinperignon, line 78: Maybe we can move all the logic of loadBody to the model
Because before this PR, loadBody was just calling a method from the viewModel

valentinperignon (approved):

PR #1110, opened 2023-11-10
fix(Threads): Merge existing thread together + Delete duplicated thread

Ambrdctr:


MailCore/Models/Thread.swift
```diff
@@ -45,6 +45,7 @@ public class Thread: Object, Decodable, Identifiable {
     @Persisted public var flagged: Bool
     @Persisted public var answered: Bool
     @Persisted public var forwarded: Bool
+    @Persisted public var folderId = ""
```
PhilippeWeidmann, line 48: We could easily create a migration to populate this field, would we benefit from it ?

MailCore/Cache/MailboxManager/MailboxManager+Message.swift
```diff
@@ -469,15 +478,51 @@ public extension MailboxManager {
...
+
+    private func removeDuplicatedThreads(
+        messageIds: MutableSet<String>,
+        threadsToUpdate: inout Set<Thread>,
+        using realm: Realm
+    ) {
+        // Create a map with all duplicated Threads of the same Thread in a list.
+        var map = [String: [Thread]]()
+        let threads = realm.objects(Thread.self).where { $0.messageIds.containsAny(in: messageIds) }
+        for thread in threads {
```
PhilippeWeidmann, line 510: I think that you could use `init(grouping:by:)`
https://developer.apple.com/documentation/swift/dictionary/init(grouping:by:)

PhilippeWeidmann (approved):

PR #1116, opened 2023-11-14
fix: Scroll loop issue on iPadOS and macOS

adrien-coye:
An attempt at fixing the "scroll to top in a loop" bug ~~that seems to be a macOS only issue~~.

EDIT: 
The issue comes from `.onAppear` not properly called _on iPad_ (so macOS as well as of today).
A quickfix is to give a proper`id` to the `MessageListView`, and it works fine. I mean no random scroll, and an expected scroll to last message of a thread, like today on iPhone.

### DIFF
--- a/Mail/Views/Thread/MessageListView.swift
+++ b/Mail/Views/Thread/MessageListView.swift
@@ -50,7 +50,7 @@ struct MessageListView: View {
                     return
                 }
 
-                // TODO: listen for last message `.isFulyLoaded` to be exact
+                // TODO: listen for last message `.isFullyLoaded` to be exact
                 DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
                     withAnimation {
                         proxy.scrollTo(firstExpandedUid, anchor: .top)
--- a/Mail/Views/Thread/MessageListView.swift
+++ b/Mail/Views/Thread/MessageListView.swift
@@ -26,8 +26,6 @@ struct MessageListView: View {
 
     @State private var messageExpansion = [String: Bool]()
 
-    @State private var scrollTask: Task<Void, Never>?
-
     var body: some View {
         ScrollViewReader { proxy in
             LazyVStack(spacing: 0) {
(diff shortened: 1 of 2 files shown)


Mail/Views/Thread/MessageListView.swift
```diff
@@ -46,16 +55,21 @@ struct MessageListView: View {
...
                     withAnimation {
-                        proxy.scrollTo(firstExpanded.uid, anchor: .top)
+                        proxy.scrollTo(firstExpandedUid, anchor: .top)
                     }
                 }
             }
             .onChange(of: messages) { newValue in
                 computeExpansion(from: newValue.toArray())
             }
+            .id(self.id)
```
adrien-coye, line 72: It works without this on iPhone as it is a different screen. It did not work on iPad.

adrien-coye: Or we can just remove the animation on iPad.

adrien-coye: Discussed

Mail/Views/Thread/ThreadView.swift
```diff
@@ -95,7 +95,7 @@ struct ThreadView: View {
                 .padding(.bottom, value: .regular)
                 .padding(.horizontal, value: .regular)
 
-                MessageListView(messages: thread.messages)
+                MessageListView(messages: thread.messages.freezeIfNeeded())
```
adrien-coye, line 98: @PhilippeWeidmann I can rollback if this is an issue.

Mail/Views/Thread/MessageListView.swift
```diff
@@ -46,16 +46,21 @@ struct MessageListView: View {
                 computeExpansion(from: messages.toArray())
 
                 guard messages.count > 1,
-                      let firstExpanded = firstExpanded() else { return }
+                      let firstExpandedUid = firstExpanded()?.uid else {
+                    return
+                }
+
+                // TODO: listen for last message `.isFulyLoaded` to be exact
```
PhilippeWeidmann, line 53:
```suggestion
                // TODO: listen for last message `.isFullyLoaded` to be exact
```

Mail/Views/Thread/ThreadView.swift
PhilippeWeidmann: @ObservedRealmObject are always frozen so it shouldn't be needed

PhilippeWeidmann, line 98: We can pass `MessageListView(messages: messages.toArray()` directly so that we do it only once (and adapt MessageListView to accept an array instead of List)

adrien-coye: Good to know thanks.

adrien-coye: Good idea, it does clean the code quite nicely.

PhilippeWeidmann (approved):

PR #1119, opened 2023-11-15
fix: Better sync calendar on macOS

adrien-coye:
### Abstract

I dynamically resolve mac types on Catalyst to open the pref pane from a specific URL.
It looks like the standard SwiftUI open URL does not work on Mac for this. (probably sec related)

### Binary size

I also had a binary size question as I needed to add the `Dynamic` package. Here is the archive size:
Archive with : 293.3 MB
Archive without : 292.1 MB

I'm happy with this, it's a non issue.

### Misc

refactor: Use .isMac where possible

### DIFF
--- a/MailCore/Utils/PlatformDetectable.swift
+++ b/MailCore/Utils/PlatformDetectable.swift
@@ -38,6 +38,7 @@ public protocol PlatformDetectable {
     var isDebug: Bool { get }
 }
 
+// TODO: This could be migrated to a Macro
 public struct PlatformDetector: PlatformDetectable {
     public init() {
         // META: Keep SonarCloud happy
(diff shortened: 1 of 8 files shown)


MailCore/Utils/PlatformDetectable.swift
```diff
@@ -38,6 +38,7 @@ public protocol PlatformDetectable {
     var isDebug: Bool { get }
 }
 
+// TODO: This could be migrated to a Macro
```
adrien-coye, line 41: TODO: remove noise

adrien-coye:
Will use `x-apple.systempreferences:com.apple.preference.security?Others_Profiles` instead

PR #1127, opened 2023-11-17
fix: Make sure we refresh signatures, one task at a time

adrien-coye:


MailCore/Cache/BackgroundRealm.swift
```diff
@@ -38,9 +38,12 @@ public final class BackgroundRealm {
 
     public func execute<T>(_ block: @escaping (Realm) -> T, completion: @escaping (T) -> Void) {
         BackgroundExecutor.executeWithBackgroundTask { [weak self] taskCompleted in
-            self?.queue.async {
-                guard let realm = self?.getRealm() else { return }
-                realm.refresh()
```
adrien-coye, line 43: No longer need a refresh here, it is done by the "generic"  getRealm now.

MailCore/Cache/BackgroundRealm.swift
```diff
@@ -38,9 +38,12 @@ public final class BackgroundRealm {
...
+            guard let self else {
+                return
+            }
+
+            queue.async {
```
PhilippeWeidmann, line 45: Do we want to keep a reference to self to execute the async block even if we try to deinit the whole object ?

adrien-coye: Correct I applied the weak / strong self pattern by reflex, as it is usualy cleaner.

Yeah I guess we can keep the double unwrap, se we can exit within the serial queue.

MailCore/Cache/ContactManager/ContactManager+Merge.swift
```diff
@@ -155,7 +159,8 @@ extension ContactManager {
         let cleanupRealm = getRealm()
         try? cleanupRealm.safeWrite {
             for emailToDelete in toDelete {
-                guard let objectToDelete = cleanupRealm.object(ofType: MergedContact.self, forPrimaryKey: emailToDelete) else {
+                guard let objectToDelete = cleanupRealm.object(ofType: MergedContact.self, forPrimaryKey: emailToDelete),
+                      !objectToDelete.isInvalidated else {
```
PhilippeWeidmann, line 163: I don't think `isInvalidated` is needed because we get the object directly from realm

adrien-coye: Good point, let me check in the documentation and change it

PR #1130, opened 2023-11-17
fix: Reduce hangs

adrien-coye:
### Abstract

Creating contacts on the fly may result in a lot of system requests to get details of a contact.

Current architecture expects to be able to build a CommonContact in main thread syncronously.
Without a more deep refactor, I was able to :

- Make the current `AvatarView` load _asyncronously_ the contact (and the underlying avatar).
- Reduce mainqueue load by building a simple cache mechanism.


MailCore/Models/CommonContact.swift
```diff
@@ -46,14 +46,117 @@ public struct AvatarImageRequest {
...
+public enum CommonContactBuilder {
+    case recipient(recipient: Recipient, contextMailboxManager: MailboxManager)
+    case user(user: UserProfile)
+    case contact(contact: CommonContact)
+    case emptyContact
+}
+
+/// Creating a CommonContact is expensive, relying on a cache to reduce hangs
+public enum CommonContactCache {
+    static let cache = NSCache<NSNumber, CommonContact>()
```
adrien-coye, line 58: Do we need one cache per mailbox ? DI can quickly solve this if it is required.

Mail/Components/AvatarView.swift
```diff
@@ -22,16 +22,34 @@ import MailResources
 import NukeUI
 import SwiftUI
 
-struct AvatarView: View, Equatable {
+struct AvatarView: View, Equatable, Identifiable {
```
PhilippeWeidmann, line 25: I understand that you marked `Identifiable` because the equality is tied to an id but this is kind of an anti-pattern as Identifiable is used for classes and structs tied to the view.

Mail/Components/AvatarView.swift
```diff
@@ -52,5 +70,6 @@ struct AvatarView: View, Equatable {
             }
         }
         .accessibilityLabel(MailResourcesStrings.Localizable.contentDescriptionUserAvatar)
+        .id(id)
```
PhilippeWeidmann, line 73: I would put directly `viewModel.displayablePerson.id` here but even then I don't think this would be right.
The id should be on the `AvatarView` to tie it to the displayed contact.

Mail/Components/AvatarView.swift
```diff
@@ -22,16 +22,34 @@ import MailResources
...
+    }
+
+    /// Equatable
+    static func == (lhs: AvatarView, rhs: AvatarView) -> Bool {
+        lhs.id == rhs.id
+    }
+
+    init(mailboxManager: MailboxManager?, contactBuilder: CommonContactBuilder, size: CGFloat = 28) {
+        self.mailboxManager = mailboxManager
+        viewModel = AvatarViewModel(contactBuilder: contactBuilder)
```
PhilippeWeidmann, line 46: `AvatarViewModel` is not a StateObject but is `Observable` this won't work. 
Additionally, if you mark it as a StateObject, the AvatarViewModel won't be recreated every time. You will have to tie AvatarView's identity to the contact. (See my next comment)
[StateObject doc about initialization](https://developer.apple.com/documentation/swiftui/stateobject#Force-reinitialization-by-changing-view-identity)

Mail/Components/AvatarView.swift
```diff
@@ -22,16 +22,34 @@ import MailResources
 import NukeUI
 import SwiftUI
 
-struct AvatarView: View, Equatable {
+struct AvatarView: View, Equatable, Identifiable {
     /// Optional as this view can be displayed from a context without a mailboxManager available
     let mailboxManager: MailboxManager?
 
-    let displayablePerson: CommonContact
+    var size: CGFloat
```
valentinperignon, line 29:
```suggestion
    let size: CGFloat
```

MailCore/Models/CommonContact.swift
```diff
@@ -46,14 +46,117 @@ public struct AvatarImageRequest {
     }
 }
 
-public struct CommonContact {
+public enum CommonContactBuilder {
```
valentinperignon, line 49: I'm afraid the name `CommonContactBuilder` might be confusing, one could imagine that it is a resultBuilder

adrien-coye: Good point, so as something that wrapps all the parameters to build a contact.
Some LLM I wil not name did propose this as names (skip the first two):

    CommonContactBuilder
    ContactBuilder
    CommonContactConstructor
    ContactParametersWrapper
    ContactSetup
    ContactConfiguration

@valentinperignon Which one do you prefer ?

MailCore/Models/CommonContact.swift
adrien-coye: Not one cache per mailbox, but a prefix in the key is probably better.

MailCore/Models/CommonContact.swift
adrien-coye: Discussed, used ContactConfiguration

Mail/Components/AvatarView.swift
adrien-coye: Ok to be more in the spirit of swiftUI I can make this ID explicit with `viewModel.displayablePerson.id`

Mail/Components/AvatarView.swift
adrien-coye: I did know that `StateObject` was to be used with `ObservableObject` but I did miss it. Thanks for the catch and the documentation.

Also none of my id/hash code is required, I played with it as it was not working for a too obvious reason and I froze.
With a proper { StateObject / ObservableObject / Published } for the view model, it works just fine.

Mail/Components/AvatarView.swift
adrien-coye: Not needed in the end.

MailCore/ViewModel/AvatarViewModel.swift
```diff
@@ -0,0 +1,63 @@
...
+@MainActor
+public final class AvatarViewModel: ObservableObject {
+    @Published public var displayablePerson: CommonContact
+
+    /// Something not using the MainActor
+    private struct AvatarLoader {
+        var contactConfiguration: ContactConfiguration
+
+        /// Async get contact method
+        func getContact() async -> CommonContact {
```
adrien-coye, line 32: Let me know if this async code is to your liking @PhilippeWeidmann 

Here I used an internal type `AvatarLoader` to escape the main actor limitation of the view model.

MailCore/Models/Contact/CommonContactCache.swift
```diff
@@ -0,0 +1,74 @@
...
+/// A standard cache for `CommonContact`, used by DI
+public typealias ContactCache = NSCache<NSNumber, CommonContact>
+
+/// Creating a `CommonContact` is expensive, relying on a cache to reduce hangs
+public enum CommonContactCache {
+    /// The underlying standard cache
+    @LazyInjectService private static var cache: ContactCache
+
+    /// Get a contact from cache if any or nil
+    public static func getContactFromCache(contactConfiguration: ContactConfiguration) -> CommonContact? {
```
adrien-coye, line 32: The public API of `CommonContactCache` is static func for clarity. But I'm open to use DI as well and use instance methods.

adrien-coye (commented):
I'm happy with this code, openning review, OK to merge asap.

PhilippeWeidmann (approved):

PR #1136, opened 2023-11-23
fix: Freeze ContactConfiguration before sending to AvatarLoader

PhilippeWeidmann:
To resolve Sentry `MAIL-IOS-5D1`

### DIFF
--- a/MailCore/ViewModel/AvatarViewModel.swift
+++ b/MailCore/ViewModel/AvatarViewModel.swift
@@ -30,7 +30,7 @@ public final class AvatarViewModel: ObservableObject {
 
         /// Async get contact method
         func getContact() async -> CommonContact {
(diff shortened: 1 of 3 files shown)


MailCore/ViewModel/AvatarViewModel.swift
```diff
@@ -30,7 +30,7 @@ public final class AvatarViewModel: ObservableObject {
 
         /// Async get contact method
         func getContact() async -> CommonContact {
-            return CommonContact.from(contactConfiguration: contactConfiguration)
+            return CommonContactCache.getOrCreateContact(contactConfiguration: contactConfiguration)
```
adrien-coye, line 33: Genuine good catch this one.

adrien-coye (approved):
Thanks for catching this.

PR #1141, opened 2023-11-24
fix: Clean message body

valentinperignon:


MailCore/Utils/SwiftSoupUtils.swift
```diff
@@ -18,17 +18,43 @@
 
 import SwiftSoup
 
-public enum SwiftSoupUtils {
-    public static func extractHTML(from document: Document, _ cssQuery: String) throws -> String {
-        guard let foundElement = try document.select(cssQuery).first() else {
-            throw SwiftSoupError.elementNotFound
+public struct SwiftSoupUtils {
```
adrien-coye, line 21: As we clean a notification, here we are fine, but what do you think of marking `functions` of this class async ?

`extractText` is already async, but it looks like this is the only one.

That way, if work is triggered from a view, it will not be run on the main queue.

adrien-coye (approved):
LGTM, I added a remark about using async functions. Let me know what you think.

MailCore/Utils/SwiftSoupUtils.swift
adrien-coye, line 1: Perfect 👌

MailCore/Utils/NotificationsHelper.swift
```diff
@@ -201,7 +201,7 @@ public enum NotificationsHelper {
         }
 
         do {
-            let cleanedDocument = try SwiftSoupUtils(from: body).cleanBody()
+            let cleanedDocument = try await SwiftSoupUtils(from: body).cleanBody()
```
adrien-coye, line 204: OK nice, we are in an async function, works as a drop-in replacement.

MailCore/Models/Draft.swift
```diff
@@ -225,7 +225,7 @@ public final class Draft: Object, Codable, Identifiable {
         return Draft(to: [recipient.detached()])
     }
 
-    public static func replyingBody(message: Message, replyMode: ReplyMode) -> String {
+    public static func replyingBody(message: Message, replyMode: ReplyMode) async -> String {
```
adrien-coye, line 228: This method is ok, as it is static and has no side effects

MailCore/Cache/DraftContentManager.swift
```diff
@@ -313,7 +313,7 @@ public class DraftContentManager: ObservableObject {
 
     private func loadReplyingMessageAndFormat(_ message: Message, replyMode: ReplyMode) async throws -> String {
         let replyingMessage = try await loadReplyingMessage(message, replyMode: replyMode)
-        return Draft.replyingBody(message: replyingMessage, replyMode: replyMode)
+        return await Draft.replyingBody(message: replyingMessage, replyMode: replyMode)
```
adrien-coye, line 316: We need to make sure `replyingMessage` is frozen or detached to keep Realm happy.
If it is, no need to add some extra code.

MailCore/Cache/DraftContentManager.swift
```diff
@@ -137,13 +137,13 @@ public class DraftContentManager: ObservableObject {
         return try await loadReplyingMessage(messageReply.message, replyMode: messageReply.replyMode).body?.freezeIfNeeded()
     }
 
-    public func replaceContent(subject: String? = nil, body: String) {
+    public func replaceContent(subject: String? = nil, body: String) async {
         guard let liveDraft = try? getLiveDraft() else { return }
-        guard let parsedMessage = try? SwiftSoup.parse(liveDraft.body) else { return }
+        guard let parsedMessage = try? await SwiftSoup.parse(liveDraft.body) else { return }
```
adrien-coye, line 142: This one gets a bit more tricky, this func becomes async, so it runs in the collaborative task pool.
Also we deal with a `liveDraft` from realm that need to be used in the same Thread.
But at this line, we now have a potential Thread change.

I would change this method tu use a frozen (or detached) object to call `SwiftSoup.parse(liveDraft.body)`

Then later when we write the new subject in a transaction L151, that's where I would get a live object.

Mail/Views/Thread/WebView/WebViewModel+SwiftSoup.swift
```diff
@@ -36,7 +36,7 @@ extension WebViewModel {
         guard let rawHTML = value else { return .errorEmptyInputValue }
 
         do {
-            guard let safeDocument = try? SwiftSoupUtils(from: rawHTML).cleanCompleteDocument()
+            guard let safeDocument = try? await SwiftSoupUtils(from: rawHTML).cleanCompleteDocument()
```
adrien-coye, line 39: All good, drop in replacement.

Mail/Views/AI Writer/Proposition/AIPropositionView.swift
adrien-coye, line 1: Using Task {} here is fine 👍 
Later we could refactor our buttons to support async functions.

adrien-coye (commented):
I have a couple of remarks, thanks for the changes 🙏

adrien-coye (approved):
All good, I did a quick manual testing pass, including draft creation / sending and also the share ext.

PR #1149, opened 2023-11-29
fix: Add more logging to watch for mailbox/folder switch

PhilippeWeidmann:


MailCore/Cache/MainViewState.swift
```diff
@@ -28,7 +28,13 @@ public class MainViewState: ObservableObject {
     ///
     /// The selected thread is the last in collection, by convention.
     @Published public var threadPath = [Thread]()
-    @Published public var selectedFolder: Folder
+    @Published public var selectedFolder: Folder {
+        didSet {
+            Task.detached { [uid = selectedFolder.remoteId, name = selectedFolder.name] in
```
adrien-coye, line 33: This works, but we might want to have a shared private method in `SentryDebug` that do a `addBreadcrumb` in a task.

MailCore/Utils/SentryDebug.swift
```diff
@@ -149,6 +149,16 @@ public enum SentryDebug {
         return crumb
     }
 
+    private static func addAsyncBreadcrumb(level: SentryLevel,
+                                           category: String,
+                                           message: String,
+                                           data: [String: Any]? = nil) {
+        Task.detached {
```
adrien-coye, line 156: ok but here as `SentryDebug` is not `mainActor` you probably do not need the `detached`

PhilippeWeidmann: Absolutely correct, will fix 👍

adrien-coye (approved):
LGTM 👍

PR #1150, opened 2023-11-29
refactor: Remove ShapeStyle from IKIcon

valentinperignon:


adrien-coye (approved):
I did a quick non regression test, all good on my side.

PR #1151, opened 2023-11-29
fix(ThreadListViewModel): Throttle sections update

PhilippeWeidmann:


Mail/Views/Thread List/ThreadListViewModel+Observation.swift
```diff
@@ -55,10 +55,8 @@ extension ThreadListViewModel {
 
                     DispatchQueue.main.sync {
                         self.filteredThreads = filteredThreads
-                        withAnimation(animateInitialThreadChanges ? .default : nil) {
-                            self.sections = newSections
-                        }
                     }
+                    sectionsSubject.send(newSections)
```
PhilippeWeidmann, line 59: Technically this is not the same as before as the call is now async and throttled but the I didn't see issues with my tests.

Mail/Views/Thread List/ThreadListViewModel+Observation.swift
```diff
@@ -41,7 +41,7 @@ extension ThreadListViewModel {
 
     // MARK: - Observe global changes
 
-    func observeChanges(animateInitialThreadChanges: Bool = false) {
+    func observeChanges() {
```
PhilippeWeidmann, line 44: Now we always perform an animation. I seems like doesn't have any side effect even at initial launch.

Mail/Views/Thread List/ThreadListViewModel.swift
```diff
@@ -182,6 +187,13 @@ final class DateSection: Identifiable, Equatable {
         self.mailboxManager = mailboxManager
         self.folder = folder
         self.isCompact = isCompact
+        sectionsObserver = sectionsSubject
+            .throttle(for: .seconds(1), scheduler: DispatchQueue.main, latest: true)
```
PhilippeWeidmann, line 191: I put a 1 sec throttle. Maybe it's too much, didn't see any issue in my tests.

adrien-coye (approved):
Looks good to me, I'll test it in beta.

valentinperignon (approved):
seems amazing

PR #1161, opened 2023-12-05
fix: Show sync DiscoveryView directly

Lnamw:


adrien-coye (approved):
Looks alright at first glance.

Mail/Views/SplitView.swift
```diff
@@ -231,7 +231,7 @@ struct SplitView: View {
             return false
         }
 
-        return appLaunchCounter.value > Constants.minimumOpeningBeforeSync
+        return true
```
valentinperignon, line 234: Maybe we can simplify the logic? Wh should show the view if `UserDefaults.shared.shouldPresentSyncDiscovery && !appLaunchCounter.isFirstLaunch`?
We might even be able to remove `appLaunchCounter.isFirstLaunch` as we want to present this bottom sheet asap

PR #1166, opened 2023-12-06
fix: Realm crash when quickly deleting attachments

adrien-coye:


PhilippeWeidmann:
This wasn't introduced by your changes but I noticed that there is a crash when quickly deleting maybe it's also solvable in this PR ?

Mail/Views/AI Writer/AIModel.swift
```diff
@@ -79,7 +79,7 @@ final class AIModel: ObservableObject {
     init(mailboxManager: MailboxManager, draftContentManager: DraftContentManager, editedDraft: EditedDraft) {
         self.mailboxManager = mailboxManager
         self.draftContentManager = draftContentManager
-        draft = editedDraft.draft.freezeIfNeeded()
+        draft = editedDraft.detachedDraft
```
adrien-coye, line 82: As the object is detached, no need for a freeze.

Mail/Views/New Message/Attachments/AttachmentsManager.swift
```diff
@@ -33,9 +33,26 @@ final class AttachmentUploadTask: ObservableObject {
     }
 }
 
-@MainActor
-final class AttachmentsManager: ObservableObject {
-    private let draft: Draft
+/// Abstract the fact some object was updated
+protocol ContentUpdatable {
```
adrien-coye, line 37: This will become more helpful in the `concurrency` PR

PR #1177, opened 2023-12-14
fix: Provide default date when decoding Thread

PhilippeWeidmann:
The threads returned by search route can sometimes return a nil date.


adrien-coye (approved):
All good.

PR #1178, opened 2023-12-15
fix(ContactManager): Do not index contacts by email

valentinperignon:
With the current ContactManager logic, we can only have one MergedContact per email address.

However, several contacts can have the same email address but a different name.
It even happens that some API contacts share the same name and email address. In this case, we only keep one MergedContact, but merge the interesting data together (such as the avatar).


PhilippeWeidmann (approved):

adrien-coye (approved):
All good 👍 

There is an issue with avatar loading, but not linked to this PR, I'll investigate.

PR #1184, opened 2023-12-20
feat(MultiSelection): Avatar enable multi selection

Ambrdctr:


Mail/Views/Thread List/ThreadListCell.swift
```diff
@@ -108,9 +115,10 @@ struct ThreadListCell: View {
     }
 
     private func didLongPressCell() {
```
valentinperignon, line 117: Can we rename this function because it no longer only applies to a long press?

Mail/Views/Thread List/ThreadListCell.swift
```diff
@@ -108,9 +115,10 @@ struct ThreadListCell: View {
...
+        let wasEnabled = multipleSelectionViewModel.isEnabled
```
valentinperignon, line 118: Can we use a `guard` here instead?

valentinperignon (approved):

PR #1185, opened 2023-12-20
fix: Show subfolders in MoveView

valentinperignon:


Mail/Views/MoveEmailView.swift
```diff
@@ -25,41 +25,56 @@ import RealmSwift
...
 
-    @Environment(\.dismissModal) var dismissModal
+    @LazyInjectService private var matomo: MatomoUtils
 
+    @Environment(\.dismissModal) private var dismissModal
     @EnvironmentObject private var actionsManager: ActionsManager
 
-    typealias MoveHandler = (Folder) -> Void
-
     @ObservedResults(Folder.self, where: { $0.role != .draft && $0.toolType == nil }) var folders
```
adrien-coye, line 35: It looks fine for a move view, but if the view is a bit unresponsive I would suggest to add a viewModel.

We could then observe the Folders with realm in a serial queue (not main) an process the various properties (filteredFolders / roleFolders / userFolders) asyncronously.

adrien-coye (approved):
LGTM 👍 

Let me know if you think the viewModel comment in necessary (it probably is not)

Mail/Views/MoveEmailView.swift
adrien-coye: Yeah that's a good point actually 👍

adrien-coye (approved):
LGTM, I have not ran the code but looks good for a beta.

PR #1187, opened 2023-12-21
feat(AvatarView): Add EmptyContact case

valentinperignon:
Useful when we have no recipient for example
Example


adrien-coye (approved):
👍

PR #1194, opened 2023-12-28
feat: Settings improved navigation

PhilippeWeidmann:
For now this only handled notifications deeplinking. However, other sections / view can easily be handled using deeplinking.

### DIFF
--- a/Mail/Views/Settings/SettingsNavigationView.swift
+++ b/Mail/Views/Settings/SettingsNavigationView.swift
@@ -23,7 +23,11 @@ import SwiftUI
 struct SettingsNavigationView: View {
     @Environment(\.dismiss) private var dismiss
 
-    @State var navigationPath: [SettingsDestination]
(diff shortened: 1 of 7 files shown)


Mail/Views/Settings/SettingsNavigationView.swift
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import MailCore
+import NavigationBackport
+import SwiftUI
+
+struct SettingsNavigationView: View {
+    @Environment(\.dismiss) private var dismiss
+
+    @State var navigationPath: [SettingsDestination]
```
valentinperignon, line 26:
```suggestion
    @State private var navigationPath: [SettingsDestination]
```

PhilippeWeidmann: It can't be private because it's init from outside. However I can refactor to make this part more understandable

valentinperignon (approved):

PR #1209, opened 2024-01-10
feat(MailError): Use `errorDescription` from LocalizedError

valentinperignon:


MailCore/Utils/Error+Extension.swift
```diff
@@ -42,7 +42,9 @@ private func displayErrorIfNeeded(error: Error) {
     @InjectService var snackbarPresenter: SnackBarPresentable
     if let error = error as? MailError {
         if error.shouldDisplay {
-            snackbarPresenter.show(message: error.errorDescription)
+            if let errorDescription = error.errorDescription {
```
PhilippeWeidmann, line 45: Maybe unwrap in same if so that we can send it to Sentry just in case ?

valentinperignon: I wondered if I should do it! I'll do this 👍

PhilippeWeidmann (approved):

PR #1212, opened 2024-01-11
feat: AuthorizationView

Lnamw:


Mail/Views/Onboarding/AuthorizationView.swift
```diff
@@ -0,0 +1,97 @@
...
+                    .padding(.top, UIPadding.onBoardingLogoTop)
+            }
+
+            VStack(spacing: UIPadding.small) {
+                Button(MailResourcesStrings.Localizable.contentDescriptionButtonNext) {
+                    if !isLastSlide {
+                        isScrollDisabled = false
+                        Task {
+                            if await (try? CNContactStore().requestAccess(for: .contacts)) != nil {
+                                try await accountManager.currentMailboxManager?.contactManager.refreshContactsAndAddressBooks()
```
valentinperignon, line 68: Is it a good thing to wait until the call is made before changing slides? We should slide to the next authorization view as soon as we have authorized contacts I think

PhilippeWeidmann:
I think users were expecting something more like this:
https://stackoverflow.com/a/33139355/9066648
However this is requires a lot more code so I'm okay to first merge this.

valentinperignon (approved):

PR #1219, opened 2024-01-12
chore: Bump ios-core

adrien-coye:


.package.resolved
PhilippeWeidmann, line 1: This shouldn't be deleted. This is either an Xcode or Tuist bug

PR #1223, opened 2024-01-16
refactor: Improved factory splitting for each target

PhilippeWeidmann:
Instead of EarlyDI hook each target has now a class extending `TargetAssembly`.
`getCommonServices` is the minimal list of all the services shared by the targets.

Each target should subclass `TargetAssembly` and override `getTargetServices` to provide additional, target related, services.

I also noticed that AppDelegate incorrectly recalled EarlyDIHook so I removed it.


adrien-coye (approved):
Yeah it was the correct time to give this a refactor, since we did use a lot the DI, which make me happy 🙌 
PS: I checked the code, I did not run it.

PR #1227, opened 2024-01-16
feat: Attachment title will be used to populate Draft Subject

adrien-coye:
*** conditions may apply, depends on type of underlying NSItemProvider.

### DIFF
--- a/Mail/Views/New Message/Attachments/Attachable.swift
+++ b/Mail/Views/New Message/Attachments/Attachable.swift
@@ -38,27 +38,32 @@ extension NSItemProvider: Attachable {
         return UTType(preferredIdentifier)
     }
 
-    public func writeToTemporaryURL() async throws -> URL {
+    public func writeToTemporaryURL() async throws -> (url: URL, title: String?) {
         switch underlyingType {
         case .isURL:
             let getPlist = try ItemProviderURLRepresentation(from: self)
-            return try await getPlist.result.get().url
+            let result = try await getPlist.result.get()
+            return (result.url, result.title)
 
         case .isText:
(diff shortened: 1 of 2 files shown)


Mail/Views/New Message/Attachments/Attachable.swift
```diff
@@ -38,27 +38,32 @@ extension NSItemProvider: Attachable {
         return UTType(preferredIdentifier)
     }
 
-    public func writeToTemporaryURL() async throws -> URL {
+    public func writeToTemporaryURL() async throws -> (url: URL, title: String?) {
```
adrien-coye, line 41: maybe a marginal improvement using a typealias for this `(url: URL, title: String?)`

MailCore/Cache/AttachmentsManagerWorker/AttachmentsManagerWorker.swift
```diff
@@ -88,6 +88,25 @@ public final class AttachmentsManagerWorker {
         liveAttachments.map { $0.detached() }
     }
 
+    var attachmentsTitles: [String?]? {
```
adrien-coye, line 91: maybe some doc

adrien-coye:
I tested the feature on iPhone8 iOS16.7 all good. Removing draft status.

PhilippeWeidmann:
I think users were expecting something more like this:
https://stackoverflow.com/a/33139355/9066648
However this is requires a lot more code so I'm okay to first merge this.

adrien-coye:
I tried to match the behaviour of Android that does not just work for webpages, therefore I did what was to me the most pragmatic first step. The good thing is that what you are sharing can be added later @PhilippeWeidmann

PR #1236, opened 2024-01-19
feat: Use background realm for MailboxInfosManager

PhilippeWeidmann:


adrien-coye (commented):
Question in DM

adrien-coye (approved):
Discussed the use of `detached` in DM. All good.

PR #1168, opened 2023-12-07
refactor: Explicit distinction between frozen and detached obj.

adrien-coye:
- First draft
- TODO check performance implications
- TODO discuss patterns with team and Android.

PR #1246, opened 2024-01-26
refactor: Explicit frozen

adrien-coye:
Successor of  https://github.com/Infomaniak/ios-kMail/pull/1168/


Mail/Views/Thread List/ThreadListViewModel.swift
```diff
@@ -174,7 +174,7 @@ final class DateSection: Identifiable, Equatable {
     ) {
         assert(folder.isFrozen, "ThreadListViewModel.folder should always be frozen")
         self.mailboxManager = mailboxManager
-        self.folder = folder
+        frozenFolder = folder.freezeIfNeeded()
```
PhilippeWeidmann, line 177: We don't need to freezeIfNeeded because of the assert at line 175

Mail/Views/Search/SearchViewModel.swift
```diff
@@ -182,8 +186,9 @@ enum SearchState {
 
         isLoading = true
         await tryOrDisplayError {
+            let searchFolderCopy = searchFolder.detached()
```
PhilippeWeidmann, line 189: Since searchFolder is already frozen we don't need to detach it

PhilippeWeidmann (changes requested):

Mail/Views/Thread List/ThreadListViewModel.swift
adrien-coye: Yup 👍 
Renaming the contructor.

adrien-coye: I did that.

Mail/Views/Search/SearchViewModel.swift
adrien-coye: I did that, and further renamed parameters where it made sense to explicit frozen state. see 59241b.

PhilippeWeidmann (approved):

PR #1273, opened 2024-02-09
feat(Search): Show folder on search result

Ambrdctr:


Mail/Views/Thread/ThreadView.swift
```diff
@@ -86,6 +86,8 @@ struct ThreadView: View {
                             ExternalRecipientView(externalTagSate: externalTag, isDraft: false)
                         }
                     }
+
+                    MessageFolderCell(title: thread.searchFolderName, inThreadHeader: true)
```
valentinperignon, line 90: If possible, MessageFolderCell (which is actually a tag) should be displayed next to the external tag. Maybe we can use a WrappingHStack

valentinperignon (approved):

PR #1277, opened 2024-02-12
fix: Improve search cache clearing

PhilippeWeidmann:


Mail/Views/Search/SearchView.swift
```diff
@@ -90,6 +90,9 @@ struct SearchView: View {
                 CloseButton {
                     Constants.globallyResignFirstResponder()
                     mainViewState.isShowingSearch = false
+                    Task {
+                        await viewModel.mailboxManager.clearSearchResults()
```
valentinperignon, line 94: Should we do it on the `.onDisappear` function?

PhilippeWeidmann: We can't because the SearchView is not presented modally

valentinperignon (approved):

PR #1289, opened 2024-02-20
feat: Show SubBody in MessageView

Ambrdctr:


Mail/Views/Thread/WebView/WebViewModel.swift
```diff
@@ -64,13 +65,36 @@ final class WebViewModel: NSObject, ObservableObject {
...
+            messageBody! += formatSubBodyContent(subBodies: subBodies, messageUid: messageUid)
+        }
+
         let loadResult = await loadHTMLString(value: messageBody, blockRemoteContent: blockRemoteContent)
         return loadResult
     }
 
+    private func formatSubBodyContent(subBodies: List<SubBody>, messageUid: String) -> String {
+        var subBodiesContent = ""
+        for subBody in subBodies {
```
adrien-coye, line 81: It is great like this. If you want to have fun and try to remove the mutable `subBodiesContent` I think you can achieve it with a `reduce`. That said, reduce can be harder to comprehend so I'm on the fence with using it.

adrien-coye (commented):
I did a review. It all looks good 👍

Mail/Views/Thread/WebView/WebViewModel.swift
Ambrdctr: That's a good idea !

PhilippeWeidmann (approved):

PR #1292, opened 2024-02-21
fix: Use email + name as primary key for MergedContacts

PhilippeWeidmann:


adrien-coye (approved):
All good. Thanks for catching this.

PR #1298, opened 2024-02-23
fix: Limit subject to 998 characters

Ambrdctr:


PhilippeWeidmann (changes requested):
We also need to truncate on send

PhilippeWeidmann (approved):

PR #1302, opened 2024-02-26
fix: Change logout message

charlella:


valentinperignon (dismissed):

MailResources/Localizable/it.lproj/Localizable.strings
```diff
@@ -1035,7 +1032,7 @@
 "recentSearchesTitle" = "Ricerche recenti";
 
 /* loco:62baabfc07e6a001ca19e535 */
-"refreshTokenError" = "Sei stato disconnesso a causa di un problema. Riconnettiti";
+"refreshTokenError" = "Siete stati disconnessi";
```
valentinperignon, line 1035: Can you fix the string on Loco and re-import it?
The correct string should be "Sei stato disconnesso"

valentinperignon (approved):

PR #1326, opened 2024-03-06
chore: Remove iOS 17.4 patch for toolbar

valentinperignon:
The betas of iOS 17.4 had an issue which caused the title of the navbar to be wrongly displayed. It was fixed in the final version.


adrien-coye (approved):
👍

PR #1331, opened 2024-03-12
chore: Add platform detector to appVersion

charlella:


MailCore/Utils/Constants.swift
```diff
@@ -161,8 +161,16 @@ public enum Constants {
         let release = Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString") as! String? ?? "x.x"
         let build = Bundle.main.object(forInfoDictionaryKey: "CFBundleVersion") as! String? ?? "x"
         let betaRelease = Bundle.main.isRunningInTestFlight ? "beta" : ""
+        let platformDetector = PlatformDetector()
```
valentinperignon, line 164: You can take a look at other files where we use PlatformDetector. 
We do not create an instance directly, instead, we use the dependency injection like this
`@InjectService private var platformDetector: PlatformDetectable`

MailCore/Utils/Constants.swift
```diff
@@ -161,8 +161,16 @@ public enum Constants {
...
+        var device = ""
```
valentinperignon, line 165: You can do it with a ternary operator. Do you see how to do it?
You can look at the variables above

The name device bothers me as it's the system name or os name

MailCore/Utils/Constants.swift
```diff
@@ -161,8 +161,16 @@ public enum Constants {
...
 
-        return "\(appName) iOS version \(release)-\(betaRelease)\(build)"
+        if platformDetector.isMac {
```
valentinperignon, line 167: Can you check if we can get the name with this property?
https://developer.apple.com/documentation/uikit/uidevice/1620054-systemname

valentinperignon (approved):

PR #1332, opened 2024-03-12
fix: Use swiftModalPresentation

Ambrdctr:


.package.resolved
```diff
@@ -1,4 +1,5 @@
 {
+  "originHash" : "c835d5365298c352cf12c2f148d73c4557fae07c9fd5e9066ffe848439b86528",
```
PhilippeWeidmann, line 2: Only commit package modifications as Ci's Xcode is not up to date

PR #1345, opened 2024-03-20
chore: Update lottie to fix onboarding animation

charlella:


PhilippeWeidmann:
Sadly performance is still bad :(

PR #1360, opened 2024-03-25
fix: Ensure UserAccountScene WindowGroup has an identifier

PhilippeWeidmann:
To test that a single window is restored simply force quit the app or stop the app from Xcode


adrien-coye (approved):
I did a quick test to test like described in this PR, looks good.

PR #1364, opened 2024-03-26
fix(WebViewModel): Instantiate configuration before WKWebView to fix datadetectors

PhilippeWeidmann:


Mail/Views/Thread/WebView/WebViewModel.swift
```diff
@@ -48,7 +48,12 @@ final class WebViewModel: NSObject, ObservableObject {
     }
 
     override init() {
-        webView = WKWebView()
+        var configuration = WKWebViewConfiguration()
```
valentinperignon, line 51:
```suggestion
        let configuration = WKWebViewConfiguration()
```
Can it be a let as WKWebViewConfiguration is a class?

PhilippeWeidmann: Yup, i thought it was a a struct

PR #1371, opened 2024-03-27
feat: Text shared via the ShareExtension is sanitized and added to a new draft

adrien-coye:
Bump swift-concurrency : https://github.com/Infomaniak/swift-concurrency/pull/4

### DIFF
--- a/Mail/Views/New Message/Attachments/AttachmentsManager.swift
+++ b/Mail/Views/New Message/Attachments/AttachmentsManager.swift
@@ -116,8 +116,8 @@ import SwiftUI
         await worker.completeUploadedAttachments()
     }
 
-    func processTextAttachments(_ attachments: [TextAttachable]) async {
-        await worker.processTextAttachments(attachments)
+    func processHtmlAttachments(_ attachments: [HTMLAttachable]) async {
+        await worker.processHtmlAttachments(attachments)
     }
 
     func attachmentUploadTaskOrFinishedTask(for uuid: String) -> AttachmentUploadTask {
--- a/Mail/Views/New Message/Attachments/AttachmentsManager.swift
+++ b/Mail/Views/New Message/Attachments/AttachmentsManager.swift
@@ -116,8 +116,8 @@ import SwiftUI
         await worker.completeUploadedAttachments()
     }
 
-    func processHtmlAttachments(_ attachments: [HTMLAttachable]) async {
-        await worker.processHtmlAttachments(attachments)
+    func processHTMLAttachments(_ attachments: [HTMLAttachable]) async {
+        await worker.processHTMLAttachments(attachments)
     }
 
     func attachmentUploadTaskOrFinishedTask(for uuid: String) -> AttachmentUploadTask {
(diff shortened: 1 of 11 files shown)


Mail/Views/New Message/Attachments/AttachmentsManager.swift
```diff
@@ -116,8 +116,8 @@ import SwiftUI
         await worker.completeUploadedAttachments()
     }
 
-    func processTextAttachments(_ attachments: [TextAttachable]) async {
-        await worker.processTextAttachments(attachments)
+    func processHtmlAttachments(_ attachments: [HTMLAttachable]) async {
```
valentinperignon, line 119: As the type is called HTMLAttachable, shouldn't the function be called processHTMLAttachments?

MailCore/Cache/Attachments/AttachmentsManagerWorker/AttachmentsManagerWorker.swift
```diff
@@ -370,22 +370,14 @@ extension AttachmentsManagerWorker: AttachmentsManagerWorkable {
         await updateDelegate?.contentWillChange()
     }
 
-    public func processTextAttachments(_ attachments: [TextAttachable]) async {
-        // Process all text attachments
-        let textAttachments = await attachments.concurrentMap { attachment in
-            await attachment.textAttachment
-        }
-
+    public func processHtmlAttachments(_ htmlAttachments: [HTMLAttachable]) async {
```
valentinperignon, line 373: I would call the function processHTMLAttachments here too (as it would follow Swift naming rules)

MailShareExtension/Attachment/WeblocToHTMLAttachment.swift
```diff
@@ -40,20 +40,39 @@ struct WeblocToTextAttachment: TextAttachable {
...
+                  let bodyUrl = URL(string: urlString) else {
+                return nil
+            }
+
+            let bodyAbsoluteUrl = bodyUrl.absoluteString
+            guard !bodyAbsoluteUrl.isEmpty else {
+                return nil
+            }
+
+            let finalHTML = "<div class=\"renderedHTML\"><a href=\"\(bodyAbsoluteUrl)\">" + bodyAbsoluteUrl + "</a></div>"
```
valentinperignon, line 74: Why do we need an HTML class renderedHTML?

adrien-coye: The idea was if we needed to debug an EML of a buggy email some day, we could see where this part of the DOM came from easily.

Mail/Views/New Message/Attachments/AttachmentsManager.swift
adrien-coye: okay sure, so HTML in caps.

MailShareExtension/Attachment/WeblocToHTMLAttachment.swift
valentinperignon: I'm not a big fan of the idea of including unnecessary attributes, rather keep the code simple

adrien-coye: Sure, I agree also. I'll remove the extraneous class.

valentinperignon (approved):

PR #1387, opened 2024-04-17
feat: Settings data privacy

Lnamw:


MailCore/Utils/MailUserDefaults+Extension.swift
```diff
@@ -199,6 +201,30 @@ public extension UserDefaults {
         }
     }
 
+    var isMatomoAuthorized: Bool {
```
adrien-coye, line 204: Nice to have, someday it would be nice to have a common read/write generic method that takes the key as argument  (like: `.matomoAuthorized`). That way we would reduce the duplication of code.

It looks like that's code duplication SonarCube is not able to detect.

adrien-coye (commented):
The code so far looks good, and I'd be ok merging this. I leave it to @PhilippeWeidmann to approve.

If you want to push further, a refactor may be useful in order to reduce code duplication.

MailCore/Utils/Logging.swift
```diff
@@ -73,7 +73,11 @@ public enum Logging {
                 #if DEBUG || TEST
                 return nil
                 #else
-                return event
+                if !UserDefaults.shared.isSentryAuthorized {
```
PhilippeWeidmann, line 76: What do you think about reversing the condition to improve readability ?

MailCore/Utils/Constants.swift
```diff
@@ -66,6 +66,7 @@ public struct URLConstants {
     public static let kdriveAppStore = URLConstants(urlString: "https://itunes.apple.com/app/id1482778676")
     public static let importMails = URLConstants(urlString: "https://import-email.infomaniak.com")
     public static let matomo = URLConstants(urlString: "https://analytics.infomaniak.com/matomo.php")
+    public static let sourceCode = URLConstants(urlString: "https://github.com/Infomaniak/ios-kMail")
```
valentinperignon, line 69: I think the name `sourceCode` is not very explicit. What do you think about something like `githubRepository`?

MailResources/Assets.xcassets/matomo.imageset/matomo.svg
valentinperignon, line 1: Would it be possible to have both light and dark icons in the same size ? 24x24?

MailResources/Localizable/it.lproj/Localizable.strings
```diff
@@ -1154,6 +1156,15 @@
 /* loco:627a53b37b420855254af032 */
 "settingsCancellationPeriodTitle" = "Periodo di cancellazione dell’invio";
 
+/* loco:6616815d2f83f25c2c0023e2 */
+"settingsDataManagementDescription" = "È possibile personalizzare i dati di utilizzo e operativi che Infomaniak recupera per correggere i bug e migliorare l’applicazione. Entrambi gli strumenti sono ospitati e gestiti esclusivamente da noi.\n\nInfomaniak rispetta la vostra privacy: il codice sorgente di questa applicazione può essere verificato, le informazioni raccolte sono anonime e i vostri dati personali rimangono strettamente confidenziali e non vengono mai condivisi con terze parti.";
+
+/* loco:661e2f0a9eed20101603dc02 */
+"settingsDataManagementSourceCode" = "Vedere il codice sorgente";
```
valentinperignon, line 1163:
```suggestion
"settingsDataManagementSourceCode" = "Vedi il codice sorgente";
```

MailResources/Assets.xcassets/data-privacy-blue.imageset/data-privacy-blue 1.svg
valentinperignon, line 1: Why both light and dark are the same?

Lnamw: I do have a really good explanation to your question, but it will be too long to write it here.

MailCore/Utils/MailUserDefaults+Extension.swift
Lnamw: I created a task on Trello so we can discuss it at our next daily

PhilippeWeidmann (approved):

PR #1394, opened 2024-04-24
fix: Improve sentry error logging

PhilippeWeidmann:
Easier to read commit by commit.

### shouldSendToSentry:
I extracted logic to choose whether we should send the events or not  -> `shouldSendToSentry`
Errors are now sent only if:
- They are not cancel errors (Native or wrapped in Alamofire)
- They are not network errors

### MailError encodable
MailErrors are now encodable to improve their readability in Sentry

### Deprecated symbols
Some deprecated Sentry calls where replaced

### Silenced warnings
Warnings regarding casting of some errors in Sentry's `context` are now handled correctly


adrien-coye (approved):
It was easy to review commit by commit. All good. Thanks 👍

PR #1397, opened 2024-04-25
chore: Bump to ios-core 10.0.0

adrien-coye:
### Abstract
Update dependencies for ios-core 10.0.0. 
Use factorised code from `core-db`, will update to transaction API in another PR.


MailCore/Cache/MailboxInfosManager.swift
```diff
@@ -123,7 +124,7 @@ public final class MailboxInfosManager {
     public func removeMailboxesFor(userId: Int) {
         let realm = getRealm()
         let userMailboxes = realm.objects(Mailbox.self).where { $0.userId == userId }
-        try? realm.uncheckedSafeWrite {
```
adrien-coye, line 126: Only the `uncheckedSafeWrite` was used in mail and this is 100% the `safeWrite` present in `core-db`

PhilippeWeidmann (approved):

PR #1401, opened 2024-04-26
ci: Detect unused code with Periphery

PhilippeWeidmann:
Unused code is now detected using periphery. (https://github.com/peripheryapp/periphery)
For every PR a comment is generated which gives a list of unused code + the count.

### DIFF
--- a/scripts/periphery.sh
+++ b/scripts/periphery.sh
@@ -24,3 +24,10 @@ unusedCount=$(wc -l <<< "$detailedOutput" | tr -d '[:space:]')
 
 echo "$detailedOutput"
 echo "Total unused instances $unusedCount"
+
+# Output script result to GITHUB_OUTPUT to get it in next step
+echo 'detailed_output<<EOF' >> $GITHUB_OUTPUT
+echo "$detailedOutput" >> $GITHUB_OUTPUT
+echo 'EOF' >> $GITHUB_OUTPUT
+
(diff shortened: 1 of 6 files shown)


scripts/periphery.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+eval "$($HOME/.local/bin/mise activate -C $SRCROOT bash --shims)"
+
+tuist install
+tuist generate -n
```
adrien-coye, line 6: yup -n makes sense

scripts/periphery.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Check if the "--full-dir" argument is provided
+if [[ "$1" == "--full-dir" ]]; then
+    full_dir=true
+else
+    full_dir=false
+fi
+
+# Perform the periphery scan and handle directory removal based on argument
+if "$full_dir"; then
```
adrien-coye, line 17: Defining a bool just to if else on it, we could probably put everything in the first parameter check.

adrien-coye (approved):
Looks good to me. Checkout my comment about defining an extra variable, you can either change it or merge this as is. As you prefer.

scripts/periphery.sh
PhilippeWeidmann: yup totally agree

valentinperignon (approved):

PR #1413, opened 2024-05-03
refactor: Substitute BackgroundExecutor for ExpiringActivity

adrien-coye:
Use `ExpiringActivity` from ios-core package instead.

The remnants of `BackgroundRealm` are then completely removed in a follow-up PR here https://github.com/Infomaniak/ios-kMail/pull/1414


MailCore/Utils/Model/Realm/BackgroundRealm.swift
```diff
@@ -38,16 +39,12 @@ public final class BackgroundRealm {
...
-                taskCompleted()
-            }
-        } onExpired: {
-            let expiredBreadcrumb = Breadcrumb(level: .warning, category: "BackgroundRealm")
-            expiredBreadcrumb.message = "Task expired before completing"
-            SentrySDK.addBreadcrumb(expiredBreadcrumb)
+        let expiringActivity = ExpiringActivity()
+        expiringActivity.start()
+        queue.async {
+            let realm = self.getRealm()
```
PhilippeWeidmann, line 45: Do we really want strong self ?

PhilippeWeidmann, line 45: Maybe something like 
```suggestion
        queue.async { [weak self] in
            guard let self else {
                 expiringActivity.endAll()
                 return
            }
            let realm = self.getRealm()
```

adrien-coye: Yes it is closer to the previous code tu use weak here.

This class goes away in the next PR so I would say, that's not an issue anyway ?

PhilippeWeidmann: As this class no longer exists in the next PR I'm resolving the conversation

PhilippeWeidmann (approved):

PR #1414, opened 2024-05-03
refactor: Remove BackgroundRealm for MailCoreRealmAccessor

adrien-coye:
Substitute `BackgroundRealm` for `Transactionable` as no longer make sense.

Should be merged after https://github.com/Infomaniak/ios-kMail/pull/1413


MailCore/Cache/MailboxManager/MailboxManager+Thread.swift
```diff
@@ -327,89 +322,82 @@ public extension MailboxManager {
...
-                            threadsToUpdate.insert(parent)
-                        }
+        guard !uids.isEmpty,
+              !Task.isCancelled else {
+            return
+        }
+
+        let batchSize = 100
+        for index in stride(from: 0, to: uids.count, by: batchSize) {
+            try? writeTransaction { writableRealm in
```
adrien-coye, line 332: Good ol switcheroo. 
The transaction is inside the for loop, and its implementation has an autoreleasepool. I'm expecting similar results memory wise as the previous solution.

MailCore/Cache/ContactManager/ContactManager.swift
```diff
@@ -85,7 +84,7 @@ public final class ContactManager: ObservableObject, ContactManageable {
                 AddressBook.self
             ]
         )
-        backgroundRealm = BackgroundRealm(configuration: realmConfiguration)
+        let backgroundRealm = MailCoreRealmAccessor(realmConfiguration: realmConfiguration)
```
adrien-coye, line 87: rename to `realmAccessor`

PhilippeWeidmann (commented):
As a general comment, now errors are throwed at an upper level. From the code I read they could also be shown to the user which I don't think we want.
Before we used `try? realm.write` so the errors were ignored. As a good middle ground, maybe we could catch and send to sentry.

adrien-coye:
@PhilippeWeidmann Sure I'll match the old pattern.

My idea was all (DB write) errors should bubble up as they are critical. Silently fails on a write error is probably not ideal.

MailCore/Cache/MailboxManager/MailboxManager+Thread.swift
PhilippeWeidmann: Before, `backgroundTracker` was preventing the app from suspending while executing the for loop. Now it could suspend in between for loop steps.

PhilippeWeidmann:
I'm see the idea but I don't think this is good UX, the user doesn't care / doesn't know what is a db write error. If we bubble up errors they have to be correctly translated and have a meaningful message.

MailCore/Cache/MailboxManager/MailboxManager+Thread.swift
adrien-coye: Yeah I'll protect the entire thing just to be sure.

adrien-coye: done

PhilippeWeidmann (approved):

PR #1416, opened 2024-05-08
fix: Remove ai and sync features from macOS

Lnamw:


PhilippeWeidmann (changes requested):
Instead of "hiding" the view it would be better to never trigger `aiModel.isShowingDiscovery` and use the platform detector where `aiModel.isShowingDiscovery` is true.

(Same goes for for the `isShowingSyncDiscovery`)

Mail/Views/Settings/SettingsView.swift
```diff
@@ -89,23 +89,27 @@ struct SettingsView: View {
...
                     }
-                    .buttonStyle(.plain)
 
                     // MARK: AI Writer
 
-                    if featureFlagsManageable.isEnabled(.aiMailComposer) {
-                        SettingsSubMenuCell(title: MailResourcesStrings.Localizable.aiPromptTitle, subtitle: aiEngine.title) {
-                            SettingsAIEngineOptionView()
+                    if !platformDetector.isMac {
+                        if featureFlagsManageable.isEnabled(.aiMailComposer) {
```
PhilippeWeidmann, line 109:
```suggestion
                    if !platformDetector.isMac && featureFlagsManageable.isEnabled(.aiMailComposer) {
```

PhilippeWeidmann: After reviewing the rest of the code I'm wondering if we could use the FeatureFlagManager to do this. 
We'll discuss it with the rest of the team.

PR #1418, opened 2024-05-13
fix(ShareExtension): UI fixes + prevent invalidated attachment crash

PhilippeWeidmann:
General improvements for the PleaseLoginView:
- Fix missing padding
- Stop using spacer
- Fix french translation

Fix crash with attachements:
- Fix for MAIL-IOS-X7Y, attachment delete button could be tapped a second time while being removed ending with an invalidated object crash.


adrien-coye (approved):
LGTM

PR #1419, opened 2024-05-13
fix: Add Empty signature in settings

Ambrdctr:


Mail/Views/Settings/Mailbox/MailboxSignatureSettingsView.swift
```diff
@@ -67,10 +75,12 @@ struct MailboxSignatureSettingsView: View {
         }
     }
 
-    private func setAsDefault(_ signature: Signature) {
-        guard !signature.isDefault else { return }
-        let detachedSignature = signature.detached()
-        detachedSignature.isDefault = true
+    private func setAsDefault(_ signature: Signature?) {
+        guard !(signature?.isDefault ?? false) else { return }
```
valentinperignon, line 79: Is it possible to do something like
```suggestion
        guard signature?.isDefault != true else { return }
```

valentinperignon (approved):

PR #1425, opened 2024-05-15
fix: Close share extension on safari mac

charlella:


Mail/Views/New Message/ComposeMessageToolbarModifier.swift
```diff
@@ -39,7 +39,7 @@ struct ComposeMessageToolbarModifier: ViewModifier {
             .navigationBarTitleDisplayMode(.inline)
             .toolbar {
                 ToolbarItem(placement: .navigationBarLeading) {
-                    if !platformDetector.isMac {
+                    if !platformDetector.isMac || Bundle.main.isExtension {
```
adrien-coye, line 42: There is a component that uses DI to do just this.
I'd prefer we use this single source of truth for now instead of checking `Bundle.` directly.

```swift
@LazyInjectService private var platformDetector: PlatformDetectable
let isInExtension = platformDetector.isInExtension
```

On kDrive you can use `AppContextService` for this. 
It is better as it does not rely on Bundle at all and knows exactly which extension is in use (share / notifications …)
We should bring it to Mail at some point, you could if you want 🙌 
```swift
@LazyInjectService var appContextService: AppContextServiceable
```

adrien-coye (changes requested):
Hey, sorry to ask for a change, but we should use our own lib instead of checking Bundle.

PhilippeWeidmann (changes requested):
Condition is also missing here `CurrentComposeMailboxView.swift:90`

PhilippeWeidmann (approved):

adrien-coye (approved):
Thanks for the changes 🙌

PR #1426, opened 2024-05-21
feat: Use InfomaniakOnboarding

PhilippeWeidmann:
This branch now uses InfomaniakOnboarding. This is the start of an effort to provide the same look and feel for onboarding across all of our applications. 

For performance and compatibility reasons InfomaniakOnboarding is written using UIKit.
Because we are using UIKit you can encounter many places where we use "glue" code to bridge between UIKit. 
Subviews that are still managed by SwiftUI are moved to their own files:
- `OnboardingBottomButtonsView`
- `OnboardingTextView`
- `OnboardingThemePickerView`
- `WaveView`

Some other notable changes:
- We can now choose if we render animations using CoreAnimation or CPU (default to CoreAnimation when possible)
- Dropped Lottie 3.5 in favor of 4.4.3. 
- Removed our custom LottieView


Mail/Views/NoMailboxView.swift
```diff
@@ -24,25 +25,17 @@ struct NoMailboxView: View {
...
-                    MailResourcesAsset.logoText.swiftUIImage
-                        .resizable()
-                        .scaledToFit()
-                        .frame(height: UIConstants.onboardingLogoHeight)
-                        .padding(.top, UIPadding.onBoardingLogoTop)
-                }
+    private let slide = Slide(backgroundImage: MailResourcesAsset.onboardingBackground3.image,
+                              backgroundImageTintColor: UserDefaults.shared.accentColor.secondary.color,
+                              content: .illustration(MailResourcesAsset.noMailbox.image),
+                              bottomViewController: UIHostingController(rootView: OnboardingTextView(
```
valentinperignon, line 32: Would it be possible to have a slide init that accepts a View?
kDrive is the only app that will provide actual UIKit. Mail and SwissTransfer will both provide View embedded in UIHostingController.

Mail/Views/Onboarding/OnboardingBottomButtonsView.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    @ModalState(context: ContextKeys.onboarding) private var isPresentingCreateAccount = false
+
+    @StateObject private var loginHandler = LoginHandler()
+
+    @Binding var selection: Int
+
+    let slideCount: Int
+
+    var isLastSlide: Bool {
```
valentinperignon, line 36:
```suggestion
    private var isLastSlide: Bool {
```

Mail/Views/Onboarding/OnboardingTextView.swift
```diff
@@ -0,0 +1,40 @@
...
+    let description: String
+
+    var body: some View {
+        VStack(spacing: 0) {
+            Text(title)
+                .textStyle(.header2)
+
+            Text(description)
+                .textStyle(.bodySecondary)
+                .padding(.top, value: .medium)
```
valentinperignon, line 32: Is it possible to remove the padding and use the spacing property of the VStack?

valentinperignon:
I sometimes have slight color issues. Maybe this issue is due to the new rendering method of Lottie.
This is a very small issue compared to the current situation.

Thanks for the onboarding library!

valentinperignon (approved):
You can merge whenever you want

PR #1438, opened 2024-05-29
fix: Do not freeze folder inside write transaction

PhilippeWeidmann:
We cannot freeze object during a write transaction. Realm will raise exception -> Cannot freeze an object in the same write transaction as it was created in.

Previously this was working because write transaction was done inside backgroundRealm.execute block.

Fix for: MAIL-IOS-1285


adrien-coye (approved):
Good catch

PR #1440, opened 2024-05-29
fix: Ensure Sentry user is set earlier to detect migration issues

PhilippeWeidmann:
Sometimes NotificationService crashes on launch due to a migration issue. 
However, Sentry userid is not set yet and we end up with a generic user id.
This PR simply moves Sentry userid before migration.


adrien-coye (approved):
Good catch, will do the same on kDrive

PR #1442, opened 2024-05-31
feat: New Window button menu

PhilippeWeidmann:


Mail/Commands/NewWindowCommand.swift
```diff
@@ -0,0 +1,45 @@
...
+    let rootViewState: RootViewState
+
+    var body: some View {
+        Button(MailResourcesStrings.Localizable.buttonNewWindow) {
+            newWindow(rootViewState: rootViewState)
+        }
+        .keyboardShortcut("n", modifiers: [.option, .command])
+    }
+
+    @MainActor func newWindow(rootViewState: RootViewState) {
```
valentinperignon, line 41: Why do we want to have the rootViewState as a function parameter, given that it is already a property of the struct?

Mail/Commands/NewWindowCommand.swift
```diff
@@ -0,0 +1,45 @@
...
+        matomo.track(eventWithCategory: .menuAction, name: "newWindow")
```
valentinperignon, line 42: Please update the Infomaniak Grids file with the new shortcut

Mail/Commands/NewWindowCommand.swift
PhilippeWeidmann: Fixed

PR #1443, opened 2024-05-31
feat(ExternalTagBottomView): Move info icon to the leading side

valentinperignon:
| Before | After |
|--------|------|
| before|after|


adrien-coye (approved):
All good, sorry for the delay.

PR #1444, opened 2024-06-04
fix(AccountManager): Use SendableProperty for currentAccount

PhilippeWeidmann:
Noticed some threading issues on Sentry that could be related to currentAccount not being thread safe. Since we now have a `SendableProperty` I made a quickwin using it. It was one of the last property not thread safe in `AccountManager`


adrien-coye (approved):
Nice idea, I did not realise the mail app had the same architecture account wise as the Drive app.

PR #1447, opened 2024-06-07
feat(Search): Multi selection added

Ambrdctr:


PhilippeWeidmann:
Transparent navigation bar bug was created because of the VStack at the root of view that was only showing a List.
- Moved all the header code to `SearchFilterHeaderView`
- Moved the List to `SearchViewList`

Now the SearchViewList is conditionally rendered within a VStack.

Tested for crashes on iOS:
15.5  (non notched)
16.0  (non notched)
16.4  (non notched)
17.5 (notched)

Ambrdctr:
Waiting for ThreadListViewModel.mailboxManager + SearchViewModel.mailboxManager as MailboxManageable

PhilippeWeidmann (approved):
All good

PR #1450, opened 2024-06-13
ci: Linting for each PR SwiftFormat

PhilippeWeidmann:
Code style comments should be minimal also code style should be applied globally. 
This workflow ensures the correct code style is applied to the whole repo.


Mail/Commands/NewMessageCommand.swift
```diff
@@ -1,6 +1,6 @@
 /*
  Infomaniak Mail - iOS App
- Copyright (C) 2024 Infomaniak Network SA
+ Copyright (C) 2022 Infomaniak Network SA
```
adrien-coye, line 3: QUID about this, shouldn't we keep the original file creation date ?

.swiftformat
```diff
@@ -82,4 +82,5 @@
 --disable andOperator,opaqueGenericParameters,preferKeyPath,redundantReturn,strongOutlets,trailingCommas,unusedArguments,wrapMultilineStatementBraces
 --enable blankLineAfterImports,blankLinesBetweenImports
 --exclude DerivedData,Derived,Tuist,Project.swift,Build,.build
+--header /*\n Infomaniak Mail - iOS App\n Copyright (C) {created.year} Infomaniak Network SA\n\n This program is free software: you can redistribute it and/or modify\n it under the terms of the GNU General Public License as published by\n the Free Software Foundation, either version 3 of the License, or\n (at your option) any later version.\n\n This program is distributed in the hope that it will be useful,\n but WITHOUT ANY WARRANTY; without even the implied warranty of\n MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the\n GNU General Public License for more details.\n\n You should have received a copy of the GNU General Public License\n along with this program. If not, see <http://www.gnu.org/licenses/>.\n */
```
adrien-coye, line 85: So we have twice the GPLV3 header in two places. Could we make one depend on the other ?
Maybe point to the existing file is swiftlint has support for it.

adrien-coye (commented):
Just a couple of questions, I'll approve quickly after.

Mail/Commands/NewMessageCommand.swift
PhilippeWeidmann: This is the original creation date for this file. A wrong creation date was probably added later.
If I do:
`git log --follow --format=%ad --date default Mail/Commands/NewMessageCommand.swift | tail -1`
I get:
`Mon May 16 10:34:07 2022 +0200`

.swiftformat
PhilippeWeidmann: This is not possible for now but I will keep an eye on this

PR #1451, opened 2024-06-14
refactor: Clean code + add swipeActions 

Ambrdctr:


Mail/Views/Search/SearchViewModel.swift
```diff
@@ -37,7 +37,17 @@ enum SearchState {
     case noResults
 }
 
-@MainActor class SearchViewModel: ObservableObject {
+final class SearchViewModel: ObservableObject, ThreadListable {
+    var frozenFolder: MailCore.Folder {
```
PhilippeWeidmann, line 41:
```suggestion
    var frozenFolder: Folder {
```

Mail/Views/Search/SearchViewModel.swift
```diff
@@ -219,11 +226,25 @@ enum SearchState {
...
+    func onTapCell(thread: Thread) {
+        selectedThread = thread
+    }
+
+    func refreshSearchIfNeeded(action: Action) {
+        guard action.refreshSearchResult else { return }
+        Task {
+            // Need to wait 500 milliseconds before reloading
+            try await Task.sleep(nanoseconds: 500_000_000)
+            print("refresh")
```
PhilippeWeidmann, line 246:
```suggestion

```

Mail/Views/Thread List/ThreadListCell.swift
```diff
@@ -101,6 +101,7 @@ struct ThreadListCell: View {
     }
 
     private func didTapCell() {
+        viewModel.addToSearchHistoryIfNeeded()
```
PhilippeWeidmann, line 104: I know it had already this name but maybe we can rename it to something more meaningful like `addCurrentSearchTermToHistoryIfNeeded()` it's more verbose but we understand it's not really related to the thread / cell being tapped.

Mail/Views/Thread List/ThreadListable.swift
```diff
@@ -0,0 +1,35 @@
...
+
+enum ThreadListableType {
+    case search
+    case threadList
+}
+
+@MainActor protocol ThreadListable {
+    var mailboxManager: MailboxManager { get }
+    var frozenFolder: Folder { get }
+    var origin: ThreadListableType { get }
```
PhilippeWeidmann, line 30: `origin` seems to be only used for matomo.
Maybe we can directly have a property `matomoEventCategory: MatomoUtils.EventCategory` declared in each model (and conforming to the protocol)

PhilippeWeidmann: Oops missed it. Maybe then we can use casting instead of needing an enum.
Something like `viewModel is SearchViewModel ? true : false` ?

PR #1456, opened 2024-06-18
fix: Small fixes for macOS

PhilippeWeidmann:
Read commit by commit:

`fix: Manually pass environmentObject to ContactActionsView` -> ensure we correctly pass the env on macOS
`fix(SelectedMailboxView): Do not depend on EnvironmentObject for mailboxmanager` -> ensure we have a mailboxManager (and the correct one) for each cell
`fix: Add MenuHeaderView to toolbar only on macOS` -> add the MenuHeaderView elements directly to the menu bar

### DIFF
--- a/Mail/Views/Menu Drawer/MenuHeaderView.swift
+++ b/Mail/Views/Menu Drawer/MenuHeaderView.swift
@@ -16,6 +16,8 @@
  along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+import InfomaniakCore
+import InfomaniakDI
 import MailCore
 import MailCoreUI
 import MailResources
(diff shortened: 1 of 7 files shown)


Mail/Views/Menu Drawer/MenuHeaderView.swift
```diff
@@ -16,40 +16,68 @@
...
-                .frame(height: UIConstants.menuDrawerLogoHeight)
-
-            Spacer()
-
-            Button {
-                mainViewState.settingsViewConfig = SettingsViewConfig(baseNavigationPath: [])
-            } label: {
-                IKIcon(MailResourcesAsset.cog, size: .large)
+        if platformDetector.isMac {
+            ZStack {}
```
valentinperignon, line 38: Is it possible to put an EmptyView() instead of a ZStack {}?

Mail/Views/Menu Drawer/MenuHeaderView.swift
```diff
@@ -16,40 +16,68 @@
...
+            .padding(.vertical, value: .regular)
+            .padding(.leading, value: .medium)
+            .padding(.trailing, value: .verySmall)
+            .background(MailResourcesAsset.backgroundSecondaryColor.swiftUIColor)
+            .clipped()
+            .shadow(color: MailResourcesAsset.menuDrawerShadowColor.swiftUIColor, radius: 1, x: 0, y: 2)
+        }
+    }
+
+    var logoImage: some View {
```
valentinperignon, line 64:
```suggestion
    private var logoImage: some View {
```

Mail/Views/Menu Drawer/MenuHeaderView.swift
```diff
@@ -16,40 +16,68 @@
...
+        MailResourcesAsset.logoText.swiftUIImage
+            .resizable()
+            .scaledToFit()
+            .frame(height: menuDrawerLogoHeight)
+    }
+
+    var settingsButton: some View {
```
valentinperignon, line 71:
```suggestion
    private var settingsButton: some View {
```

Mail/Views/Menu Drawer/MenuHeaderView.swift
PhilippeWeidmann: If use EmptyView, toolbar seems to be ignored

Ambrdctr (approved):

PR #1460, opened 2024-06-18
fix: Emit a Sentry event on Token auth error

adrien-coye:
Now tracking old and new token during an update to track further issues.

Drive PR https://github.com/Infomaniak/ios-kDrive/pull/1201

### DIFF
--- a/MailCore/Utils/SentryDebug.swift
+++ b/MailCore/Utils/SentryDebug.swift
@@ -190,16 +190,6 @@ public enum SentryDebug {
         return crumb
     }
 
-    private static func addAsyncBreadcrumb(level: SentryLevel,
-                                           category: String,
-                                           message: String,
-                                           data: [String: Any]? = nil) {
-        Task {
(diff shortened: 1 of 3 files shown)


MailCore/Utils/SentryDebug.swift
```diff
@@ -343,3 +344,64 @@ public enum SentryDebug {
...
+    }
+
+    static func capture(
+        message: String,
+        context: [String: Any]? = nil,
+        contextKey: String? = nil,
+        level: SentryLevel? = nil,
+        extras: [String: Any]? = nil
+    ) {
+        Task {
```
PhilippeWeidmann, line 391: We have to be careful about what we send as `context`. If it's a realm object it will crash

adrien-coye: I merged this code with `addAsyncBreadcrumb`, and made explicit the async nature of the new methods.

PhilippeWeidmann (approved):

PR #1462, opened 2024-06-19
chore: Add strip symbols script

PhilippeWeidmann:
I've been stripping manually symbols for several release.
This does it automatically for release builds.


adrien-coye (approved):
Perfect 🤝

PR #1475, opened 2024-07-01
feat: Display Bimi

Lnamw:
- Display Bimi authentication checkmark and image
- If recipient is in contacts use Avatar image if has one
- Check if FeatureFlag is enabled

Simulator Screenshot - iPhone 15 Pro - 2024-07-01 at 15 58 05

Simulator Screenshot - iPhone 15 Pro - 2024-07-01 at 15 56 18


Mail/Components/ThreadCell/ThreadCell.swift
```diff
@@ -77,9 +77,13 @@ struct ThreadCellDataHolder {
         }
     }
 
-    func contactConfiguration(contextMailboxManager: MailboxManager) -> ContactConfiguration {
+    func contactConfiguration(thread: Thread, contextMailboxManager: MailboxManager) -> ContactConfiguration {
```
PhilippeWeidmann, line 80: Maybe pass directly bimi instead of the whole thread since no other property is used

PhilippeWeidmann (commented):
A migration is missing, simply bumping realm version in MailboxManager should be enough

Mail/Views/Thread/Message/MessageHeader/MessageHeaderDetailView.swift
```diff
@@ -80,11 +81,19 @@ struct RecipientLabel: View {
...
     @State private var contactViewRecipient: Recipient?
 
     @LazyInjectService private var matomo: MatomoUtils
 
+    init(labelWidth: Binding<CGFloat>, title: String, recipients: RealmSwift.List<Recipient>, bimi: Bimi? = nil) {
+        _labelWidth = labelWidth
+        self.title = title
+        self.recipients = recipients
+        self.bimi = bimi
+    }
```
PhilippeWeidmann, line 95: It's okay to use a var for `bimi` instead of adding a complete init

PhilippeWeidmann (approved):

PR #1484, opened 2024-07-10
fix: Message worker updates view accordingly

adrien-coye:
- ~~PoC of a fix, discussed with team about the best way to do it that is still broken for now.~~
- Refactor message worker as an ObservableObject.
- Fix issues related to view not been updated when the `InlineMessageWorker` fires a change.
- Optimized loading of base64 images. Now parallelized.
- Optimized body loading with images. Now parallelized compact and mail body processing.
- [iOS17] static images are compressed in HEIC to speed up webview rendering.
- Changed a `Thread.sleep()` for a more liberal `Thread.yield()`
  - Greatly reduce wait time for inline image loading without sacrificing device responsiveness.
- Each batch processing of inline images now has an autoreleasepool. 
- Checked if using a Mutable String internally would benefit memory usage, did not notice such an impact.
  - Did not move forward with this.

### DIFF
--- a/Mail/Views/Thread/Message/MessageView.swift
+++ b/Mail/Views/Thread/Message/MessageView.swift
@@ -49,12 +49,12 @@ struct MessageView: View {
 
     @State private var isShowingErrorLoading = false
 
-    /// Something to preprocess inline attachments
-    @State var inlineAttachmentWorker: InlineAttachmentWorker?
-
     @State var displayContentBlockedActionView = false
 
     @ObservedRealmObject var message: Message
+
+    /// Something to preprocess inline attachments
+    let inlineAttachmentWorker: InlineAttachmentWorker
 
     private var isRemoteContentBlocked: Bool {
         return (UserDefaults.shared.displayExternalContent == .askMe || message.folder?.role == .spam)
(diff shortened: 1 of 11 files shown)


Mail/Views/Thread/Message/MessageView.swift
```diff
@@ -104,19 +105,31 @@ struct MessageView: View {
             .task {
                 await fetchMessageAndEventCalendar()
             }
+            .task {
+                for await _ in inlineAttachmentWorker.objectWillChange.values {
```
adrien-coye, line 109: This is a hack, just for demo purpose

adrien-coye: gone as the wind 💨

MailCore/Utils/Realm+Extensions.swift
```diff
@@ -38,17 +38,4 @@ public extension Object {
 
         return transactionable.fetchObject(ofType: Self.self, forPrimaryKey: primaryKeyValue)
     }
-
-    /// Get an updated frozen copy for a given object
-    func refresh() -> Self {
```
adrien-coye: @Ambrdctr saw this as unused, since the code is `transactionable` oriented, I removed this.

Ambrdctr (approved):

Ambrdctr:
LGTM!

PR #1489, opened 2024-07-17
fix: Further image rendering compression on legacy hardware

adrien-coye:
On iOS17, we compress inline images with HEIC to reduce memory usage and speed up the rendering of the mail, as images in a base64 form within an HTML is not fast.

This PR could further speed-up the rendering of emails on legacy device but _at the cost of image quality_
Therefore, this is only a draft to discuss.

### DIFF
--- a/Mail/Views/Thread/Message/InlineAttachmentWorker.swift
+++ b/Mail/Views/Thread/Message/InlineAttachmentWorker.swift
@@ -238,21 +238,41 @@ struct BodyImageProcessor {
 
     /// Try to compress the attachment with the best matched algorithm. Trade CPU cycles to reduce render time and memory usage.
     private func compressedBase64ImageAndMime(attachmentData: Data, attachmentMime: String) -> ImageBase64AndMime {
-        guard #available(iOS 17.0, *) else {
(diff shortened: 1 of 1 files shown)


Mail/Views/Thread/Message/InlineAttachmentWorker.swift
```diff
@@ -238,21 +238,41 @@ struct BodyImageProcessor {
...
-        return ImageBase64AndMime(base64String, "image/heic")
+        // On ancient devices, we rely on jpg compression
+        // to accelerate rendering and reduce memory usage at the cost of some quality.
+        else {
+            let image = UIImage(data: attachmentData)
+
+            // On large files (> 5M) we do not see artifacting, even zooming within the webview
+            let compressionQuality: CGFloat
+            if attachmentData.count > 5_000_000 {
+                compressionQuality = 0.1
```
adrien-coye, line 262: To see a difference in loading times, I really need to push it. But is does speed things up on the rendering side, while not been that much visible on large images.

adrien-coye:
Supper seeded by AV1 support. https://github.com/Infomaniak/ios-kMail/pull/1492

PR #1496, opened 2024-07-23
fix: Download attachments for Mac

Lnamw:
This PR fix the bug where user couldn't save an attachement on MacOS.
It has been decided to have one "share button" on iOS and one "share button" and one "download button" on MacOS.

### MacOS
Capture d’écran 2024-07-23 à 15 48 49

### iOS
Capture d’écran 2024-07-23 à 15 51 38

### DIFF
--- a/Mail/Views/Attachment/AttachmentPreview.swift
+++ b/Mail/Views/Attachment/AttachmentPreview.swift
@@ -114,21 +114,23 @@ struct AttachmentPreview: View {
 
                     Spacer()
 
-                    Button {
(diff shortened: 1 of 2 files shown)


Infomaniak MailDebug.entitlements
PhilippeWeidmann, line 1: This looks weird. We already have an entitlements file for macOS but it doesn't end with debug

PhilippeWeidmann:
I notice that we still have an "Open in kDrive" button. Maybe we can also hide it with this PR ?

PhilippeWeidmann (approved):

PR #1502, opened 2024-08-07
fix: Reset folder if empty

Ambrdctr:
When moving more than a page of messages in a folder, If the folder was empty before, we won't be able to fetch all messages.


adrien-coye (approved):
I did take time to run it on iPhone 8. Looks good to me.

PR #1505, opened 2024-08-08
feat: Thread algo updated

Ambrdctr:
- New way to get old messages
- Get new messages using Delta


Mail/Components/Custom Buttons/LoadMoreButton.swift
PhilippeWeidmann, line 1: Could we refactor the whole button action to a function now that there is more code ?

PhilippeWeidmann (approved):

PR #1521, opened 2024-08-30
refactor: Use UserProfile instead of Account

PhilippeWeidmann:
Depends on #1520


adrien-coye (approved):
Code looks good, have not run the code tho.

PR #1531, opened 2024-09-10
fix: Cache profiles

PhilippeWeidmann:
Depends on #1521


Mail/Views/New Message/Select Mailbox/SelectComposeMailboxViewModel.swift
```diff
@@ -48,12 +48,17 @@ final class SelectComposeMailboxViewModel: ObservableObject {
     }
 
     func listProfiles() {
```
adrien-coye, line 50: It looks like the previous version of this method was syncronous. It now hides an async task. Can this be an issue if called from SwUI (glitches …)?

adrien-coye (commented):
I only reviewed the code. I did not try to run it.

Mail/Views/New Message/Select Mailbox/SelectComposeMailboxViewModel.swift
PhilippeWeidmann: Good catch. I ensured we call back on main thread

adrien-coye (approved):
All good 👍

PR #1540, opened 2024-09-12
feat: Save a mail in kDrive

Matthieu-dgl:


Matthieu-dgl:
There's still one problem: hiding the button in the mail application on Mac

MailCore/Cache/Actions/ActionsManager.swift
```diff
@@ -197,6 +199,17 @@ public class ActionsManager: ObservableObject {
                     )
                 }
             }
+        case .saveMailInkDrive:
+            Task { @MainActor in
+                if !platformDetector.isMac {
```
PhilippeWeidmann, line 204: Use guard

MailCore/Cache/Actions/Action+List.swift
```diff
@@ -93,6 +96,10 @@ extension Action: CaseIterable {
             .shareMailLink,
             userIsStaff ? .reportDisplayProblem : nil
         ]
+        if !platformDetector.isMac {
+            tempListActions.insert(.saveMailInkDrive, at: 6)
+        }
```
PhilippeWeidmann, line 101: Use ternary with nil for macOS

PR #1550, opened 2024-09-20
fix: Expose InfomaniakDI as Framework

PhilippeWeidmann:
As seen with tests on previous Xcode versions, Xcode 16 seems to also link differently Previews. With our current tuist configuration this leads to duplicate classes which causes problems with the DI singleton.


valentinperignon (approved):
🎉

PR #1560, opened 2024-09-30
fix: Fix the background color of ActionsView

Matthieu-dgl:


valentinperignon (changes requested):
Can we remove the 1.5x scale?

valentinperignon (approved):

PR #1563, opened 2024-10-01
feat: Preview in context menu

lebojo:


PhilippeWeidmann:
Sadly no solution works correctly to size the view. We are closing it since there is no way to do it properly.

PR #1580, opened 2024-10-14
fix: Async contact search bar in a new mail

Matthieu-dgl:


MailCore/Cache/ContactManager/ContactManager+DB.swift
```diff
@@ -60,11 +62,18 @@ public extension ContactManager {
             .freeze()
 
         var sortedIfNecessary: any Collection<MergedContact> = lazyResults
-        if let sorted {
-            sortedIfNecessary = sortedIfNecessary.sorted(by: sorted)
+        if let sortClosure = sorted {
```
PhilippeWeidmann, line 65: This should work ?
```suggestion
        if let sorted {
```

PhilippeWeidmann (approved):

PR #1583, opened 2024-10-15
fix(SearchView): Replace Group with VStack

valentinperignon:


PhilippeWeidmann (approved):
Tested on iOS 15.5 and 16.4 with no issue.

PR #1588, opened 2024-10-18
fix: Prevent accessing wrong offset when removing contact

PhilippeWeidmann:


Ambrdctr (approved):

Ambrdctr:
Tried to run it, did not run.
lgtm

PR #1603, opened 2024-10-31
feat(Autocompletion): Add Address books

valentinperignon:


valentinperignon:
Should be opened later by @Matthieu-dgl

PR #1606, opened 2024-11-01
fix: Multiple selection shouldn't navigate

PhilippeWeidmann:
- Removed `tapGesture` in favor of `Button` to better align with what SwiftUI expects.
- Workaround the navigation issue by disabling the multiple selection in the next run loop. (Basically added a DispatchQueue.main.async 🙃)  

Tested in simulator on iOS:
- 18.1
- 17.5
- 16.4
- 15.5

I'm open to better ideas instead of the workaround.


valentinperignon (approved):
Thank you for the fix!

PR #1608, opened 2024-11-04
fix: Lint exclude derivedData file

Matthieu-dgl:


.swiftlint.yml
```diff
@@ -38,3 +38,4 @@ excluded:
  - DerivedData/
  - Derived/
  - Tuist/.build/
+ - Tuist/DerivedData
```
adrien-coye, line 41:
```suggestion
 - Tuist/DerivedData/
```
I works without it I'm sure, but it matches the rest of the file this way.

PR #1609, opened 2024-11-04
chore: Bump dependencies

Matthieu-dgl:


valentinperignon (approved):

valentinperignon (changes requested):
We need to fix the floating panel on iPhone

valentinperignon (approved):

PR #1610, opened 2024-11-07
fix(iOS 18.1): Set explicit arrow edge position for popovers

valentinperignon:
This is necessary to fix an issue introduced by Apple on iOS 18.1 and fixed on iOS 18.2 (135231043)


adrien-coye (approved):
I did some test on old devices. All looks good to me.

PR #1633, opened 2025-01-07
chore: Bump to ios-login built in swift6 mode

adrien-coye:


valentinperignon (approved):

adrien-coye:
Just did a non regression test, I can log in the app with the new lib.

PR #1634, opened 2025-01-08
fix: Only refresh mailboxes with displayed notifications

PhilippeWeidmann:


MailCore/Cache/RefreshAppBackgroundTask.swift
```diff
@@ -64,18 +65,53 @@ public class RefreshAppBackgroundTask: BackgroundTaskable {
     }
 
     public func run() async {
-        guard let currentMailboxManager = accountManager.currentMailboxManager,
-              let inboxFolder = currentMailboxManager.getFolder(with: .inbox),
-              inboxFolder.cursor != nil else {
-            // We do nothing if we don't have an initial cursor
+        guard !accountManager.accounts.isEmpty else {
```
adrien-coye, line 68: Debatable but I prefer this. What do you think ?
```suggestion
        guard accountManager.currentAccount != nil else {
```

MailCore/Cache/RefreshAppBackgroundTask.swift
```diff
@@ -64,18 +65,53 @@ public class RefreshAppBackgroundTask: BackgroundTaskable {
...
             BGTaskScheduler.shared.cancelAllTaskRequests()
             return
         }
 
-        // Every time we execute a task we schedule a new task for next time
+        let notificationCenter = UNUserNotificationCenter.current()
+        let deliveredNotifications = await notificationCenter.deliveredNotifications()
+
+        var uniqueMailboxObjectIds = Set<String>()
+        for notification in deliveredNotifications {
```
adrien-coye, line 77: It's a taste thing but I'd map `deliveredNotifications` and cast it with`Set()` to remove mutability.

MailCore/Cache/RefreshAppBackgroundTask.swift
```diff
@@ -64,18 +65,53 @@ public class RefreshAppBackgroundTask: BackgroundTaskable {
...
+        guard !mailboxManagersToRefresh.isEmpty else {
+            BGTaskScheduler.shared.cancelAllTaskRequests()
+            return
+        }
+
+        // Only reschedule if we have at least one mailbox correctly init
         scheduleForBackgroundLaunch()
 
-        await currentMailboxManager.refreshFolderContent(inboxFolder.freezeIfNeeded())
+        await withTaskGroup(of: Void.self) { group in
```
adrien-coye, line 108: Here I would argue we could add value by using `concurrentForEach()` instead from _InfomaniakConcurrency_.
The degree of parallelism in maintained by the abstraction and scales.
Also I would say it is easier to read than a `withTaskGroup`

adrien-coye (commented):
Have a look at the notes and let me know what you think.
I'll test the code after you get back to me.

MailCore/Cache/RefreshAppBackgroundTask.swift
adrien-coye: Okay, so if `AccountManageable` encapsulates the management of user accounts, it should expose a `someUserAuthenticated` so less leaks to this component.
I feel I could make your own argument with checking the array of accounts directly also.
I'm ok with this, resolving now.

adrien-coye (approved):
All good for me.  I did test it for a while.
It looks like it does not work in Lckdwn mode, but that's beyond our control.

PR #1635, opened 2025-01-08
ci: Increase waiting delay MailUITests

PhilippeWeidmann:


adrien-coye (approved):
Make sense for _UI_ tests 👍

PR #1644, opened 2025-01-29
chore: Move settings data management to core UI

Matthieu-dgl:
To test this PR you need the host version of core UI, you need to change URL in you package
link of the core UI PR https://github.com/Infomaniak/ios-core-ui/pull/74


valentinperignon (changes requested):
Updating the matomo preference has no effect, we do not opt out matomo

PR #1648, opened 2025-02-05
fix: Lazy env for Sentry

PhilippeWeidmann:
Apple warns us about "slow code path on launch" one of them is checking if we are in TestFlight while launching.
This resolution is now lazy and only executed when sending events


adrien-coye (approved):
Code LGTM. I also ran it quickly. Approving now.

PR #1684, opened 2025-02-27
fix: Realm embedded object reuse crash

adrien-coye:


MailCore/Cache/MailboxManager/MailboxManager.swift
```diff
@@ -167,15 +167,14 @@ public final class MailboxManager: ObservableObject, MailboxManageable {
...
         }
         if keepProperties.contains(.localSafeDisplay) {
             message.localSafeDisplay = savedMessage.localSafeDisplay
         }
         if keepProperties.contains(.attachments) {
-            for attachment in savedMessage.attachments {
-                message.attachments.append(Attachment(value: attachment.freeze()))
-            }
+            let attachments = savedMessage.attachments.map { Attachment(value: $0.freeze()) }
+            message.attachments.append(objectsIn: attachments)
```
PhilippeWeidmann, line 177: I don't see how `SomeRealmObject(value: frozenObject)` could fix an issue in this context.

If we had a crash, freezing here will already be too late.

MailCore/Cache/MailboxManager/MailboxManager+Thread.swift
```diff
@@ -528,8 +528,9 @@ public extension MailboxManager {
                 return
             }
 
+            let resultThreads = result.threads?.compactMap { Thread(value: $0) } ?? []
```
PhilippeWeidmann, line 531:
```suggestion
            let resultThreads = result.threads?.map { Thread(value: $0) } ?? []
```

MailCore/Cache/MailboxManager/MailboxManager.swift
PhilippeWeidmann: Okay I see, attachment was already frozen. Still, I don't think we need to freeze body

adrien-coye: Ok, so we could also remove the other freeze in other properties then

adrien-coye: I only did a rollback of my change.

PhilippeWeidmann (approved):

PR #1689, opened 2025-03-05
fix: Show not inline files

lebojo:
Display non inline attachments files even with contentId


PhilippeWeidmann (changes requested):
This is a computed property so this will be on the main thread.

What we should do is do this computation once when we get the message here: `MailboxManager+Message.swift:26` (and also update the corresponding thread)

MailCore/Models/Attachment.swift
```diff
@@ -34,6 +34,7 @@ public class Attachment: /* Hashable, */ EmbeddedObject, Codable, Identifiable {
     @Persisted(originProperty: "attachments") var parentLink: LinkingObjects<Message>
     @Persisted public var saved = false
     @Persisted public var temporaryLocalUrl: String?
+    @Persisted public var isInline: Bool
```
PhilippeWeidmann, line 37: `isInline` needs a default value

PhilippeWeidmann (approved):

PR #1693, opened 2025-03-07
feat: Sort autocompletion result

BaptGrv:


PhilippeWeidmann (changes requested):
You need to add contacted_times in the with=

MailCore/Models/Contact/Contact.swift
```diff
@@ -74,5 +76,10 @@ public struct InfomaniakContact: Codable, Identifiable {
         nickname = try values.decodeIfPresent(String.self, forKey: .nickname)
         organization = try values.decodeIfPresent(String.self, forKey: .organization)
         groupContactId = try values.decodeIfPresent(List<Int>.self, forKey: .groupContactId)
+        if let rawContactedTimes = try values.decodeIfPresent([String: Int].self, forKey: .contactedTimes) {
```
PhilippeWeidmann, line 79: Just in case
```suggestion
        if let rawContactedTimes = try? values.decodeIfPresent([String: Int].self, forKey: .contactedTimes) {
```

PhilippeWeidmann (approved):

PR #1697, opened 2025-03-11
chore: Only run swiftlint on main target

PhilippeWeidmann:


adrien-coye (approved):
Did some tests. LGTM.

PR #1714, opened 2025-03-19
chore: Bump DI to the version using an internal lock

adrien-coye:
Having a discrete PR for this is a good reference point to track if any issue and/or performance improvements from the Lock based DI.


adrien-coye (commented):
I did test it on device, and it ran just fine. Performance wise it did feel the same.

PhilippeWeidmann (approved):

PR #1730, opened 2025-03-27
fix: Correctly observe drop success

PhilippeWeidmann:
Some kind of retain cycle was created because of the closure. Observing this way is also more correct.


lebojo (approved):
Nice! 🎉

PR #1740, opened 2025-04-01
feat: Keep attachment in cache

Ambrdctr:


MailCore/Cache/MailboxManager/MailboxManager+Message.swift
```diff
@@ -49,7 +49,14 @@ public extension MailboxManager {
...
 
-        let data = try await apiFetcher.attachment(attachment: attachment, progressObserver: progressObserver)
+        let attachmentCacheHelper = AttachmentCacheHelper()
+        let data: Data
+        if let cachedData = attachmentCacheHelper.getCache(resource: attachment.resource) {
+            data = cachedData
+        } else {
+            data = try await apiFetcher.attachment(attachment: attachment, progressObserver: progressObserver)
+            attachmentCacheHelper.storeCache(resource: attachment.resource, data: data)
+        }
```
PhilippeWeidmann, line 59: We can also move the realm write transaction in the else. We don't need to update it if it is already in cache.

Mail/MailApp.swift
```diff
@@ -35,6 +36,7 @@ struct MailApp: App {
         Logging.resetAppForUITestsIfNeeded()
         Logger.general.info("Application starting in foreground ? \(UIApplication.shared.applicationState != .background)")
         refreshAppBackgroundTask.register()
+        ImagePipeline.shared = ImagePipeline(configuration: .withDataCache)
```
PhilippeWeidmann, line 39: According to my tests it is not needed.

PhilippeWeidmann (commented):
Two small remarks and we can merge

PhilippeWeidmann (approved):

PR #1743, opened 2025-04-03
fix: Realm crash save search threads

adrien-coye:


adrien-coye (commented):
I did some tests, nothing to report. I also left some remarks for reviewers

MailCore/Cache/MailboxManager/MailboxManager.swift
```diff
@@ -180,26 +180,28 @@ public final class MailboxManager: ObservableObject, MailboxManageable {
...
         }
-        message.inTrash = savedMessage.inTrash
+        liveMessage.inTrash = savedMessage.inTrash
         if keepProperties.contains(.fullyDownloaded) {
-            message.fullyDownloaded = savedMessage.fullyDownloaded
+            liveMessage.fullyDownloaded = savedMessage.fullyDownloaded
         }
         if keepProperties.contains(.body), let body = savedMessage.body {
-            message.body = Body(value: body)
+            liveMessage.body = Body(value: body).detached()
```
PhilippeWeidmann, line 195: No need to recreate from Body from value. We can create a new one with detached.
```suggestion
            liveMessage.body = body.detached()
```

MailCore/Cache/MailboxManager/MailboxManager.swift
```diff
@@ -180,26 +180,28 @@ public final class MailboxManager: ObservableObject, MailboxManageable {
...
         }
         if keepProperties.contains(.localSafeDisplay) {
-            message.localSafeDisplay = savedMessage.localSafeDisplay
+            liveMessage.localSafeDisplay = savedMessage.localSafeDisplay
         }
         if keepProperties.contains(.attachments) {
             let attachments = savedMessage.attachments.map { Attachment(value: $0.freeze()) }
```
PhilippeWeidmann, line 201: I know you didn't touch this line but it would make more sense to use detached. Freeze isn't useful in this case.
```suggestion
            let attachments = savedMessage.attachments.map { $0.detached() }
```

adrien-coye: Sure

Also I'll keep an eye for this since it exists elsewhere in the app. We blindly use this `Type(value: )` at other locations

PhilippeWeidmann (approved):

PR #1758, opened 2025-04-11
fix: Open draft with correct content

PhilippeWeidmann:


adrien-coye (approved):
I did play with this code. It does fix the issue in my testing. LGTM.

PR #1760, opened 2025-04-14
fix: Remove useless properties mailbox

PhilippeWeidmann:


adrien-coye (approved):
LGTM, I did some tests migrating accounts and performing some login /logouts.

PR #1761, opened 2025-04-15
fix: Drop realm on migration error in production

PhilippeWeidmann:


adrien-coye (approved):
Code LGTM. 
Also tested the code with multiple accounts on failed migrations.
The execution does interrupt (as long as a debugger is hooked in debug).
Seamless to the user, I'm approving now.

PR #1762, opened 2025-04-16
fix: Drop current folder if we have too many changes

PhilippeWeidmann:


MailCore/Cache/MailboxManager/MailboxManager+Thread.swift
```diff
@@ -591,6 +604,21 @@ public extension MailboxManager {
         writableRealm.delete(orphanMessages)
     }
 
+    private func resetFolder(_ folder: Folder) async throws {
+        try writeTransaction { realm in
+            guard let liveFolder = folder.fresh(using: realm) else { return }
+
+            liveFolder.remainingOldMessagesToFetch = Constants.messageQuantityLimit
+            liveFolder.oldMessagesUidsToFetch.removeAll()
```
PhilippeWeidmann, line 612: I didn't use Folder.resetHistoryInfo because I want this block to be explicit with the properties it resets.

MailCore/Cache/MailboxManager/MailboxManager+Thread.swift
```diff
@@ -591,6 +604,21 @@ public extension MailboxManager {
...
+            liveFolder.newMessagesUidsToFetch.removeAll()
+            realm.delete(liveFolder.threads)
+            realm.delete(liveFolder.messages)
```
PhilippeWeidmann, line 615: From my tests this correctly removes the threads and messages in the db (not only those in the list)

Ambrdctr (approved):

valentinperignon (approved):

PR #1764, opened 2025-04-17
feat: Can't collapse single message

Matthieu-dgl:


Mail/Views/Thread/Message/MessageView.swift
```diff
@@ -50,13 +52,20 @@ struct MessageView: View {
...
+    }
+
     var body: some View {
         VStack(spacing: 0) {
             MessageHeaderView(
                 message: message,
                 isMessageExpanded: Binding(get: {
                     isMessageExpanded
                 }, set: { newValue in
+                    guard !isSingleMessage else { return }
```
PhilippeWeidmann, line 68: Maybe we can compute it on the fly ?
```suggestion
                    guard threadForcedExpansion.count > 1 else { return }
```

PhilippeWeidmann (approved):

PR #1775, opened 2025-04-24
fix: Hide role folder in menuDrawer tree

Ambrdctr:


MailCore/Models/NestableFolder.swift
```diff
@@ -31,6 +31,19 @@ public struct NestableFolder: Identifiable {
...
+            return true
+        }
+        for child in children {
+            if child.hasSubFolders {
+                return true
+            }
+        }
+        return false
+    }
+
```
valentinperignon, line 46: I'm a bit worried about performance with the recursive algorithm :/
It would be nice to have @PhilippeWeidmann's opinion!

Maybe we could store and compute this value in Realm?

PhilippeWeidmann: Yep, we can do it in the background when fetching the folders `MailboxManager+Folders.swift:27`

PhilippeWeidmann (approved):

PR #1794, opened 2025-05-08
fix: Use background instead of contentShape

Ambrdctr:


PhilippeWeidmann:
I didn't know that we needed the contentShape for iPad. We'll keep it

PR #1814, opened 2025-05-19
feat(EmojiReaction): Add reaction buttons UI

valentinperignon:


MailCore/Models/Recipient.swift
```diff
@@ -53,14 +57,18 @@ public final class RecipientsList: EmbeddedObject, Encodable {
         self.recipients = recipients.toRealmList()
     }
 
+    public func append(recipients: [Recipient]) {
+        self.recipients.append(objectsIn: recipients)
+    }
+
+    public func contains(where precidate: (Recipient) -> Bool) -> Bool {
+        return recipients.contains(where: precidate)
```
PhilippeWeidmann, line 65: ups
```suggestion
        return recipients.contains(where: predicate)
```

PR #1829, opened 2025-06-04
fix: Add background color on AlertView

Matthieu-dgl:
Depends on https://github.com/Infomaniak/ios-core-ui/pull/97
Depends on https://github.com/Infomaniak/ios-features/pull/17


valentinperignon (changes requested):
As discussed together 👍 
- We should avoid repeating each time the background color
- Should we also change the background color of the iPhone version of the discovery view?

valentinperignon (approved):

PR #1835, opened 2025-06-06
fix: Add matomo to messages actions

BaptGrv:


Mail/Views/Bottom sheets/Actions/ActionsView.swift
```diff
@@ -137,6 +138,8 @@ struct QuickActionView: View {
 }
 
 struct MessageActionView: View {
+    @LazyInjectService private var matomo: MatomoUtils
+
```
PhilippeWeidmann, line 142: Remove lazy and move in call site scope

valentinperignon (commented):
- Matomo events are missing for quick actions
- Events should use matomo.trackBulkEvent if it is used from multiple selection

MailCore/Utils/Matomo+Extension.swift
```diff
@@ -119,4 +119,24 @@ public extension MatomoUtils {
             track(eventWithCategory: .userInfo, action: .data, name: "multipleMessagesInThread", value: messagesCount)
         }
     }
+
+    func trackAction(action: Action, origin: ActionOrigin, numberOfItems: Int, isMultipleSelection: Bool) {
```
valentinperignon: Maybe a name more explicit would be a good idea? Something like `trackBottomSheetAction`?

Mail/Views/Bottom sheets/Actions/ActionsView.swift
```diff
@@ -203,3 +211,37 @@ struct ActionButtonLabel: View {
...
+            matomo.trackAction(
+                action: request.action,
+                origin: request.origin,
+                numberOfItems: request.targetMessages.count,
+                isMultipleSelection: request.isMultipleSelection
+            )
+        }
+    }
+}
```
valentinperignon: Not a big fan of this struct. Even if Sonar doesn't agree, having a function with 5 parameters makes more sense.

Having a top level function like `performActionAndTrack` is not a good practice.

PR #1847, opened 2025-06-19
fix: Close realm after notification processing

PhilippeWeidmann:


adrien-coye (approved):
LGTM, I did test that notifications are still OK on device.

PR #1866, opened 2025-07-04
fix: Hide mobile toolbar on macOS

PhilippeWeidmann:


adrien-coye (approved):
I did a test on MacOS. LGTM.

PR #1871, opened 2025-07-15
feat(Encryption): Concerned recipient list

Ambrdctr:


Mail/Views/Encryption/EncryptionConcernedRecipientsView.swift
```diff
@@ -16,15 +16,33 @@
...
 
 struct EncryptionConcernedRecipientsView: View {
+    @EnvironmentObject private var mailboxManager: MailboxManager
+
+    let message: Message
+
     var body: some View {
-        // TODO: - Get recipient list
-        Text("Recipient list")
+        VStack {
```
valentinperignon, line 29: Can we specify the spacing here? Guess it's something like 0?

valentinperignon (commented):
I can present the view even if I have 0 recipients concerned

valentinperignon (approved):

PR #1891, opened 2025-08-07
test: Remove unknown folders from test generator

PhilippeWeidmann:


adrien-coye (approved):
This bug was not trivial to understand, and why it was making a test flaky.
Thanks for finding this one.

PR #1894, opened 2025-08-11
chore: Release build 1.5.1-b37

PhilippeWeidmann:


adrien-coye (approved):
matches what I got on kDrive for ios-core and ios-features

PR #1927, opened 2025-08-26
feat: Hide reaction messages from search if emoji feature is available

BaptGrv:


PhilippeWeidmann (changes requested):
You need to hide it at the store level not only in the view.
This is only needed for offline threads not online since the back-end already returns the correct threads.
What you are looking for is here:
MailboxManager+Search.swift:104

MailCore/Cache/MailboxManager/MailboxManager+Search.swift
```diff
@@ -151,27 +153,29 @@ public extension MailboxManager {
...
-                    hasAttachments: newMessage.hasAttachments,
-                    hasDrafts: newMessage.isDraft,
-                    flagged: newMessage.flagged,
-                    answered: newMessage.answered,
-                    forwarded: newMessage.forwarded
-                )
-                newThread.makeFromSearch(using: writableRealm)
-                newThread.subject = message.subject
-                searchFolder.threads.insert(newThread)
+                if !(featureAvailableProvider.isAvailable(.emojiReaction) && message.emojiReaction != nil) {
```
PhilippeWeidmann, line 156: This can probably be done with an initial predicate
`MailboxManager+Search.swift:115`

PhilippeWeidmann (approved):

PR #1931, opened 2025-09-01
feat: New notification preview for an emoji reaction

BaptGrv:


MailCore/Utils/NotificationsHelper.swift
```diff
@@ -306,6 +306,20 @@ public enum NotificationsHelper {
     }
 
     private static func getCleanBodyFrom(message: Message) async -> String {
+        if let emojiReaction = message.emojiReaction {
+            let name: String
+            if let senderName = message.from.first?.name, !senderName.isEmpty {
```
valentinperignon, line 311: You can add a guard to get the `message.from.first`.
if this value is empty, you can return nil

valentinperignon (changes requested):
Nice work 👍 

It would be nice to compute this preview for the ThreadList (in case the message is in a different folder than INBOX or notifications are disabled).
You can extract this logic in a function. You can ask me when you begin working on it (it could be a formatter)

PhilippeWeidmann (approved):

PR #1933, opened 2025-09-01
feat: Reply to an encrypted mail is now encrypted too

BaptGrv:


valentinperignon (changes requested):
You can move this logic in the function `Draft.replying(reply, currentMailboxEmail)` in `Draft.swift:263)`

valentinperignon (approved):

PR #1980, opened 2025-10-01
fix: Adapt PDF preview to the size of the screen on iOS18

BaptGrv:


Mail/Views/Thread/Message/Attachment/AttachmentsView.swift
```diff
@@ -117,8 +117,14 @@ struct AttachmentsView: View {
...
-            AttachmentPreview(attachment: previewedAttachment)
-                .environmentObject(mailboxManager)
+            if #available(iOS 18.0, *) {
+                AttachmentPreview(attachment: previewedAttachment)
+                    .environmentObject(mailboxManager)
+                    .presentationSizing(.page)
+            } else {
+                AttachmentPreview(attachment: previewedAttachment)
+                    .environmentObject(mailboxManager)
+            }
```
PhilippeWeidmann, line 127: AttachmentPreview is also displayed on other screens. You can refactor to a view extension that does nothing < iOS 18.
For example: `Toolbar+Extension.swift:33`

PhilippeWeidmann (approved):

PR #2119, opened 2026-03-05
fix: Use asset for CloseButton

Ambrdctr:


valentinperignon (commented):
Not sure to understand why we want to do this 🤔

valentinperignon (approved):

PR #2134, opened 2026-03-31
feat: Add possibility to translate a message

BaptGrv:
Depends on #2168


PhilippeWeidmann (commented):
Looking good 🚀 

A few things to make it better:
- Can you add a withAnimation on banner appear / disappear
- We can do a special case for the banner to display the animated `EuriaAnimationView`
- Use the translate icon from DS/Manager instead of the magic wand

valentinperignon (changes requested):
Great :hype:

I would be nice to handle the error `translation__target_same_as_source` in a banner instead of a snackbar

The feature lacks some "withAnimation" when we display or hide the banners

Currently the "translated" option is displayed in the ThreadList and it's not correct

valentinperignon (changes requested):
We should munge the mail once it get translated as the translation version may break the layout

valentinperignon:
The API has been updated and is ready to be implemented 👍

valentinperignon (approved):

PR #2149, opened 2026-04-23
fix(TxtToHTMLAttachment): Try to unarchive the file before decoding the string

valentinperignon:


adrien-coye (approved):
Code looks good. It does fix the issue in my testing.

PR #2154, opened 2026-04-28
feat: Force mail in light

Ambrdctr:


MailResources/Assets.xcassets/view-in-light.imageset/view-in-light.svg
valentinperignon, line 1: Maybe we can use the "theme-tel" asset?

MailCore/Models/Message.swift
```diff
@@ -195,6 +201,8 @@ public final class Message: Object, Decodable, ObjectKeyIdentifiable {
     @Persisted public var reactions: List<MessageReaction>
     @Persisted public var reactionMessages: List<Message>
 
+    @Persisted public var theme: MessageTheme = .auto
```
valentinperignon: I would not store it into Realm

MailCore/Models/Message.swift
```diff
@@ -123,6 +123,12 @@ public final class MessageReaction: EmbeddedObject {
     }
 }
 
+public enum MessageTheme: String, PersistableEnum {
+    case light
+    case dark
+    case auto = "light dark"
```
valentinperignon: I would prefer a computed property `cssProperty: String` instead of a String rawValue

valentinperignon (changes requested):

MailCore/Models/Message.swift
PhilippeWeidmann: We can now use the ThreadViewState 🎉

MailResources/Assets.xcassets/view-in-light.imageset/view-in-light.svg
Ambrdctr: Looks better with "view-in-light" in my opinion

Mail/Views/Thread/Message/MessageBody/MessageBodyContentView.swift
```diff
@@ -25,17 +25,36 @@ import SwiftSoup
...
+    ) {
+        _displayContentBlockedActionView = displayContentBlockedActionView
+        _initialContentLoading = initialContentLoading
+        self.presentableBody = presentableBody
+        self.blockRemoteContent = blockRemoteContent
+        self.messageUid = messageUid
+
+        let viewModel = WebViewModel()
+        viewModel.theme = messageTheme
+        _model = StateObject(wrappedValue: viewModel)
```
PhilippeWeidmann: Aie StateObject 101
https://developer.apple.com/documentation/swiftui/stateobject#Initialize-state-objects-using-external-data

MailCore/Cache/ThreadViewState.swift
```diff
@@ -21,6 +21,9 @@ import SwiftUI
 public class ThreadViewState: ObservableObject {
     @Published public var summaries = [String: MessageSummaryState]()
     @Published public var translatedMessages = [String: MessageTranslatedState]()
+    @Published public var forcedLightModes = Set<String>()
+
+    @Published public var colorScheme: ColorScheme = .light
```
PhilippeWeidmann: Storing and forwarding the colorScheme in ThreadViewState feels wrong. 
It's more work but also more correct to pass @Environment to Action+List.swift‎.

It is not that bad if there are too many arguments we are doing a refactor of the action list creation anyway.

PhilippeWeidmann (changes requested):

PhilippeWeidmann (approved):

valentinperignon (approved):

PR #2163, opened 2026-05-04
feat: Using TargetAssembly from ios-core lib

killianmathias:


MailCore/Utils/MailAppTargetAssembly.swift
adrien-coye: This helper exists in core now. I think you can remove it safely.

MailCore/Utils/MailAppTargetAssembly.swift
```diff
@@ -49,18 +50,18 @@ open class TargetAssembly {
...
 
-    public init() {
+    override public init() {
         // Setup debug stack early
         Logging.initLogging()
         ApiEnvironment.current = Self.apiEnvironment
 
         // setup DI ASAP
-        Self.setupDI()
+        TargetAssembly.setupDI()
```
adrien-coye: I think this can be removed and replace for a standard super.init() call early that should do the same.

MailCore/Utils/MailAppTargetAssembly.swift
```diff
@@ -49,18 +50,18 @@ open class TargetAssembly {
         accessType: nil
     )
 
-    public init() {
+    override public init() {
         // Setup debug stack early
```
adrien-coye, line 48: `super.init()`

adrien-coye (commented):
Just some small remarks. Please rebase on core once merged.

adrien-coye (commented):
Looking good. Please use the released 18.10.1 ios-core and I'll merge it.

adrien-coye (approved):
Okay so code is GO and App is GO. I did test send /rec email, the share ext, drafts, contacts.

PR #2181, opened 2026-05-21
feat: Show sending message

Ambrdctr:


Mail/Views/Thread/Message/MessageHeader/MessageHeaderSummaryView.swift
```diff
@@ -108,15 +108,21 @@ struct MessageHeaderSummaryView: View {
...
                         .iconSize(.large)
                 }
                 .foregroundStyle(MailResourcesAsset.redColor)
             }
 
-            if isMessageExpanded && isMessageInteractive && !(message.isScheduledDraft ?? false) {
+            if message.isDraft && message.scheduled {
+                Text(MailResourcesStrings.Localizable.messageIsSending)
+                    .textStyle(.bodySmallSecondary)
+                    .italic()
```
valentinperignon: Maybe it can be its own TextStyle

MailCore/Utils/Constants.swift
```diff
@@ -91,6 +91,9 @@ public enum Constants {
     public static let maxAttachmentsSize = 26_214_400 // 25 mo
     public static let minimumQuotasProgressionToDisplay = 0.03
 
+    /// Delay for some API call
+    public static let gracePeriod = 2
```
valentinperignon, line 95: I wonder if we can move this constant in the DraftManager or something like this?
If not, that's okay!

valentinperignon (commented):
Really nice! Just a few comments.

Also when I click on a sending draft on the Draft folder, it opens the compose view. I guess it should not.

Can you synchronize with Android to ensure we have the same behavior?

MailCore/Utils/Constants.swift
Ambrdctr: It is a constant. To me it make more sense to have it in the Constants file.

Mail/Views/Thread/Message/MessageHeader/MessageHeaderSummaryView.swift
Ambrdctr: The textStyle already exist, I didn't see it.
I'll fix to use the correct textStyle.

PhilippeWeidmann (approved):

PR #2182, opened 2026-05-21
feat: Show inline attachments in the editor

killianmathias:


Mail/Views/Thread/Message/BodyImageProcessor.swift
```diff
@@ -113,6 +114,35 @@ struct BodyImageProcessor {
...
+    func addContentIdAttributesInBody(body: String?, attachments: [Attachment]) async -> String? {
+        guard let body, !body.isEmpty else {
+            return nil
+        }
+
+        guard let htmlBody = try? await SwiftSoup.parse(body), let imageElements = try? await htmlBody.select("img[src]") else {
+            return nil
+        }
+
+        for attachment in attachments {
```
valentinperignon: The algorithm feels a bit strange. Maybe we can just loop over the images, then add the contentID. Let's discuss it together if you want!

valentinperignon (changes requested):
The DraftContentManager loads the complete body of the draft. We have to ensure the draft is fully loaded before trying to replace its content.
We should probably move this logic to the DraftContentManager

valentinperignon (approved):

PR #2185, opened 2026-05-22
fix: Disable mail link sharing for free myKSuite users

killianmathias:


PhilippeWeidmann (changes requested):
We already have access to isShowingMyKSuiteUpgrade/isShowingKSuiteProUpgrade in MainViewState from ActionsManager. We don't need to pass it.

In ActionsManager we can do mainViewState.isShowingMyKSuiteUpgrade

PhilippeWeidmann (approved):

PR #2208, opened 2026-06-11
fix: Missing environment and DI

BaptGrv:


MailCore/Utils/Error+Extension.swift
```diff
@@ -39,9 +39,10 @@ public func tryOrDisplayError(_ body: () async throws -> Void) async {
 }
 
 private func displayErrorIfNeeded(error: Error) {
-    @InjectService var snackbarPresenter: IKSnackBarPresentable
+    guard !Bundle.main.isExtension else { return }
```
PhilippeWeidmann: I think that we have a IKSnackBarPresentable that does nothing that can be injected in extensions

PhilippeWeidmann (approved):

PR #2229, opened 2026-06-29
feat: Reminder and schedule toggles

Lnamw:


MailResources/Localizable/fr.lproj/Localizable.strings
```diff
@@ -373,6 +403,9 @@
 /* loco:62b1c0de630e5a572542c032 */
 "buttonCreateFolder" = "Ajouter un dossier";
 
+/* loco:6a327bedaf69f744ea0b9752 */
+"buttonCustomReminder" = "Délais personnalisé";
+
```
killianmathias, line 408: This string should be updated (Already fixed in Loco by Elouan)

killianmathias (approved):

PR #2258, opened 2026-07-22
fix: Fetch when app `willEnterForeground`

Muuushuuu:


Mail/UserAccountScene.swift
```diff
@@ -77,6 +78,13 @@ struct UserAccountScene: Scene {
             reviewManager.decreaseActionUntilReview()
         }
         checkAppVersion()
+
+        if case .mainView(_, let mainViewState) = rootViewState.state {
```
valentinperignon: Can probably be done on the SplitView.swift

valentinperignon (changes requested):

Mail/UserAccountScene.swift
adrien-coye: I advised this location since we already do work here (eg. `appLockHelper.startObservation()`)

adrien-coye (commented):
I did test it on my iPhone, it does work when the app is alive and jumps to foreground but not on a fresh start.

EDIT: It does not work any more on my phone when the code is moved to the split view

Mail/UserAccountScene.swift
adrien-coye: It does not work for me when called from the spliview for some reason https://github.com/Infomaniak/ios-kMail/pull/2258#pullrequestreview-4755581566 (guard prevent execution)

valentinperignon (approved):

adrien-coye (approved):
I did verify on my phone the last change it works well on startup or switch to foreground.

Mail/UserAccountScene.swift
adrien-coye: I did verify latest patch, all good on my side now

PR #2268, opened 2026-08-03
fix: Improve ContentBlocker rules to prevent unwanted remote content loading

PhilippeWeidmann:


adrien-coye (approved):
Works for me. Issues of the UI not warning external content blocked by the ".*" wilcard is out of scope. All good to me 👍

PR #2285, opened 2026-08-18
feat: Add custom long press action

Muuushuuu:
Depends On #https://github.com/Infomaniak/swift-rich-html-editor/pull/41
Depends On #2277

### DIFF
--- a/Mail/Views/New Message/ComposeMessageBodyView.swift
+++ b/Mail/Views/New Message/ComposeMessageBodyView.swift
@@ -73,16 +73,20 @@ struct ComposeMessageBodyView: View {
         return MessageWebViewUtils.loadCSS(for: .editor(aliases: aliases)).joined()
     }
 
-    @State private var commands: [HTMLEditorCustomAction] = [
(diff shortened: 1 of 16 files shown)


Mail/Views/New Message/ComposeMessageBodyView.swift
```diff
@@ -71,6 +71,17 @@ struct ComposeMessageBodyView: View {
...
+    @State private var commands: [HTMLEditorCustomAction] = [
+        HTMLEditorCustomAction(title: MailResourcesStrings.Localizable.pasteAdjustStyleAction) { editor in
+            let pasteboard = UIPasteboard.general
+            guard pasteboard.numberOfItems > 0 else { return }
+
+            if let text = pasteboard.string, !text.isEmpty {
+                editor.pasteAndMatchStyle(text)
+            }
+        }
+    ]
```
valentinperignon: If we change the way the id of HTMLEditorCustomAction is computed, it doesn't need to be a State anymore

